PR #2, opened 2022-12-16
add some logging

forwardpointer:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Add some logging details.

### Checklist


lib/uwb/UwbPeer.cxx
```diff
@@ -32,6 +52,17 @@ UwbPeer::operator=(const UwbPeer& other)
     return *this;
 }
 
+std::string
+UwbPeer::ToString() const
+{
+    std::stringstream ss;
+    ss << "{" << std::endl;
+    ss << "mac: " << m_address.ToString() << std::endl;
+    ss << m_spatialProperties.ToString() << std::endl;
```
abeltrano, line 61: If you include `<notstd/tostring.hxx>` and do a `using namespace strings::ostream_operators`, you can just pass the object directly, and it will invoke the `ToString()` method automatically:

```suggestion
    ss << "mac: " << m_address << std::endl;
    ss << m_spatialProperties << std::endl;
```

lib/uwb/UwbPeer.cxx
```diff
@@ -1,10 +1,30 @@
...
 #include <tuple>
 
+#include <sstream>
+#include <string>
 #include <uwb/UwbPeer.hxx>
-
 using namespace uwb;
 
+std::string
+UwbPeerSpatialProperties::ToString() const
```
abeltrano, line 10:
```suggestion
UwbPeerSpatialProperties::ToString() const noexcept
```

lib/uwb/UwbPeer.cxx
```diff
@@ -1,10 +1,30 @@
...
+{
+    std::initializer_list<std::tuple<std::string, std::optional<double>>> name2value{ { "Distance", Distance }, 
+                                                                                    { "AngleAzimuth", AngleAzimuth }, 
+                                                                                    { "AngleElevation", AngleElevation }, 
+                                                                                    { "Elevation", Elevation } };
+
+    std::stringstream ss;
```
abeltrano, line 17:
```suggestion
    std::ostringstream ss;
```

tools/cli/NearObjectCliUwbSessionEventCallbacks.cxx
```diff
@@ -6,31 +6,40 @@
 using namespace nearobject::cli;
 
 void
-NearObjectCliUwbSessionEventCallbacks::OnSessionEnded(::uwb::UwbSession* /* session */, ::uwb::UwbSessionEndReason /* reason */)
+NearObjectCliUwbSessionEventCallbacks::OnSessionEnded(::uwb::UwbSession* session, ::uwb::UwbSessionEndReason reason)
 {
-    std::cout << "Session Ended" << std::endl;
+    std::cout << "Session " << session->GetId() << ": Session Ended" << std::endl;
```
abeltrano, line 11: I think it would be helpful to include a prefix indicating the number is a session id. It might also help to output it with a fixed width (8 characters, since that's the max size), and as hex. Eg.

```
"Session with id=0xaabbccdd ended"
```

For controlling the width and output base of numbers, see the `UwbMacAddress` parsing code.

abeltrano (approved):
Some minor suggestions, but overall looks good.  I appreciate the size of this PR being small 👍

lib/uwb/UwbPeer.cxx
forwardpointer: isn't it possible for the string allocation to fail?

abeltrano: Yes, good call, ignore.

lib/uwb/UwbPeer.cxx
abeltrano: Since this doesn't seem to be working in this instance, let's skip it for now.

tools/cli/NearObjectCliUwbSessionEventCallbacks.cxx
```diff
@@ -1,36 +1,54 @@
...
 using namespace nearobject::cli;
+using namespace strings::ostream_operators;
 
 void
-NearObjectCliUwbSessionEventCallbacks::OnSessionEnded(::uwb::UwbSession* /* session */, ::uwb::UwbSessionEndReason /* reason */)
+NearObjectCliUwbSessionEventCallbacks::OnSessionEnded(::uwb::UwbSession* session, ::uwb::UwbSessionEndReason reason)
 {
-    std::cout << "Session Ended" << std::endl;
+    std::cout << "Session with id="
+              << "0x" << std::hex << std::setw(8) << std::setfill('0') << session->GetId() << ": Session Ended" << std::endl;
```
abeltrano, line 14: Instead of using `0x` directly, we should use the generic stream manipulator `showbase` instead. You'll also have to set the fill positioning with either `std::<left|right|internal>`. Here, `std::internal` makes sense since it will add the fill characters in between the base  (0x) and the value, eg. `0x******VV` where `*` is the fill char and `VV` is the value.

```suggestion
              << std::showbase << std::internal << std::hex << std::setw(8) << std::setfill('0') << session->GetId() << ": Session Ended" << std::endl;

```

abeltrano:
Please apply clang-format to the changed files.

abeltrano (approved):

PR #14, opened 2022-12-19
Add support for StaticRangingInfo via cli

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Adds support for entering StaticRangingInfo via cli

### Checklist

### DIFF
--- a/lib/uwb/protocols/fira/StaticRangingInfo.cxx
+++ b/lib/uwb/protocols/fira/StaticRangingInfo.cxx
@@ -1,0 +1,18 @@
+
+#include <uwb/protocols/fira/StaticRangingInfo.hxx>
+#include <sstream>
+#include <iomanip>
+
+std::string
+uwb::protocol::fira::StaticRangingInfo::ToString() const {
+    std::ostringstream ss;
+    ss << "{" << std::endl;
+    ss << "VendorId: " << VendorId << std::endl;
+    ss << "InitializationVector: " << std::hex << std::setw(2) << std::setfill('0') << std::showbase << std::internal;
+    for(const auto i: InitializationVector) {
+        ss << i << " ";
+    }
+    ss << std::endl;
+    ss << "}" << std::endl;
+    return ss.str();
+}
--- a/lib/uwb/protocols/fira/StaticRangingInfo.cxx
+++ b/lib/uwb/protocols/fira/StaticRangingInfo.cxx
@@ -7,13 +7,11 @@ std::string
 uwb::protocol::fira::StaticRangingInfo::ToString() const
 {
     std::ostringstream ss;
-    ss << "{" << std::endl;
     ss << "VendorId: " << VendorId << std::endl;
     ss << "InitializationVector: " << std::hex << std::setw(2) << std::setfill('0') << std::showbase << std::internal;
     for (const auto i : InitializationVector) {
         ss << +i << " ";
     }
     ss << std::endl;
-    ss << "}" << std::endl;
     return ss.str();
 }
--- a/lib/uwb/protocols/fira/StaticRangingInfo.cxx
+++ b/lib/uwb/protocols/fira/StaticRangingInfo.cxx
@@ -1,10 +1,11 @@
 
-#include <uwb/protocols/fira/StaticRangingInfo.hxx>
-#include <sstream>
 #include <iomanip>
+#include <sstream>
+#include <uwb/protocols/fira/StaticRangingInfo.hxx>
 
 std::string
-uwb::protocol::fira::StaticRangingInfo::ToString() const {
+uwb::protocol::fira::StaticRangingInfo::ToString() const
+{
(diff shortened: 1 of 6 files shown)


lib/uwb/protocols/fira/StaticRangingInfo.cxx
```diff
@@ -0,0 +1,19 @@
...
+{
+    std::ostringstream ss;
+    ss << "{" << std::endl;
+    ss << "VendorId: " << VendorId << std::endl;
+    ss << "InitializationVector: " << std::hex << std::setw(2) << std::setfill('0') << std::showbase << std::internal;
+    for (const auto i : InitializationVector) {
+        ss << +i << " ";
+    }
+    ss << std::endl;
+    ss << "}" << std::endl;
```
abeltrano, line 17: I'd prefer that these separators are only added for cases where there is structure nesting. This way when it's printed on its own, the caller can choose to insert their own separator chars between major structures if they so choose.

```suggestion
    ss << "VendorId: " << VendorId << std::endl;
    ss << "InitializationVector: " << std::hex << std::setw(2) << std::setfill('0') << std::showbase << std::internal;
    for (const auto i : InitializationVector) {
        ss << +i << " ";
    }
    ss << std::endl;
```

tools/cli/NearObjectCli.cxx
```diff
@@ -176,24 +178,24 @@ NearObjectCli::AddSubcommandUwbRange(CLI::App *parent)
...
+    rangeStartApp->add_option("--Channel", m_cliData->SessionData.uwbConfiguration._channel)->transform(CLI::CheckedTransformer(::detail::ChannelMap))->capture_default_str();
+    rangeStartApp->add_option("--StsPacketConfiguration", m_cliData->SessionData.uwbConfiguration._rframeConfig)->transform(CLI::CheckedTransformer(::detail::StsPacketConfigurationMap))->capture_default_str();
+    rangeStartApp->add_option("--ConvolutionalCodeConstraintLength", m_cliData->SessionData.uwbConfiguration._convolutionalCodeConstraintLength)->transform(CLI::CheckedTransformer(::detail::ConvolutionalCodeConstraintLengthMap))->capture_default_str();
+    rangeStartApp->add_option("--PrfMode", m_cliData->SessionData.uwbConfiguration._prfMode)->transform(CLI::CheckedTransformer(::detail::PrfModeMap))->capture_default_str();
+    rangeStartApp->add_option("--UwbMacAddressFcsType", m_cliData->SessionData.uwbConfiguration._macAddressFcsType)->transform(CLI::CheckedTransformer(::detail::UwbMacAddressFcsTypeMap))->capture_default_str();
```
abeltrano, line 198: We should really fix this at the source. Since we anticipate users `#include`'ing the helper namespace, it probbaly shouldn't have `detail` namespace exposed. 

I'll send a PR.

tools/cli/NearObjectCli.cxx
```diff
@@ -254,14 +260,17 @@ NearObjectCli::AddSubcommandUwbRangeStart(CLI::App *parent)
...
-            auto result = uwb::protocol::fira::StringToResultReportConfiguration(m_cliData->ResultReportConfigurationString, detail::ResultReportConfigurationMap);
+            auto result = uwb::protocol::fira::StringToResultReportConfiguration(m_cliData->ResultReportConfigurationString, ::detail::ResultReportConfigurationMap);
             if (not result) {
                 std::cout << "could not parse ResultReportConfiguration" << std::endl;
             } else {
                 m_cliData->SessionData.uwbConfiguration._resultReportConfigurations = result.value();
             }
         }
 
+        m_cliData->SessionData.staticRangingInfo = m_cliData->StaticRanging;
```
abeltrano, line 271: Is there any reason not to use `m_cliData->SessionData.staticRangingInfo` as the `assignTo` variable directly instead of having this extra level of indirection and copying?

abeltrano (dismissed):
Approved pending merge of PR #16

tools/cli/NearObjectCli.cxx
forwardpointer: I wasn't sure if you could write to a std::optional that is a std::nullopt but I will try it

forwardpointer: Just verified that you can't directly write to a member of a class wrapped in a std::optional if it is not allocated. But maybe I can initialize the std::optional as a default initialized StaticRangingInfo and then write to those members?

abeltrano: Ahh yes, I see that the member itself is a `std::optional` and CLI11 doesn't support any indirection.

abeltrano (approved):

PR #18, opened 2022-12-20
UwbConfiguration improvements

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Restore coding style in `UwbConfiguration` class and improve its handling of CLI inputs.

### Technical Details

* Introduce staging structure to populate `UwbConfiguration` objects from CLI input.
* Remove `UwbConfiguration` public members and explicit `std::optional` members. Replace with parameter <-> value map.
    - The map allows users of the class to loop through set values without having to do an explicit if check for each member. 
    - The accessors were kept for pure convenience reasons; it remains to be seen whether there is a use for them or not.
* Update `UwbConfiguration` accessor functions to resolve the `std::any` types and return `std::optional` representing whether the values were populated or not.
* Rename `RangingMethod` to `RangingDirection` and then `RangingConfiguration` to `RangingMethod` to match FiRa spec naming.
* Update `UwbConfiguration` to contain a set of parameter tag and value paris.
checks are only sacrificed for the CLI input where they can't really be avoided.

### Test Results

All unit tests pass on both Linux and Windows.

### Reviewer Focus

Consider whether there is a cleaner way to do this.

### Future Work

None

### Checklist


lib/uwb/protocols/fira/UwbConfiguration.cxx
```diff
@@ -34,179 +34,226 @@ UwbConfiguration::FromDataObject(const encoding::TlvBer& /* tlv */)
...
+}
+
+const std::unordered_map<UwbConfiguration::ParameterTag, std::any>&
+UwbConfiguration::GetValueMap() const noexcept
+{
+    return m_values;
+}
+
+bool
+UwbConfiguration::operator==(const UwbConfiguration& other) const noexcept
```
forwardpointer, line 224: does the std::unordered_map specialization for operator== work here instead?

```suggestion
return m_values == other.m_values;
```

forwardpointer: Looks like the template matching can't automatically compare std::any types

abeltrano: The stored type within `std::any` is erased, so the implementation can't determine which `operator==` overload to use, unfortunately. Hence this abomination of checks :'(

PR #22, opened 2022-12-23
Configure impl

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

ConfigureImpl implementation.

### Checklist


windows/devices/uwb/UwbSession.cxx
```diff
@@ -14,12 +16,152 @@ UwbSession::UwbSession(std::weak_ptr<uwb::UwbSessionEventCallbacks> callbacks, w
     m_handleDriver(std::move(handleDriver))
 {}
 
+template <class T>
+std::unique_ptr<UWB_APP_CONFIG_PARAM>
+GetUWB_APP_CONFIG_PARAM(const T &value)
+{
+    auto pdata = std::make_unique<uint8_t>(FIELD_OFFSET(UWB_APP_CONFIG_PARAM, paramValue[sizeof T]));
+}
```
abeltrano, line 24: This needs to use the array overload (`std::make_unique<uint8_t[]>`). Also, `std::unique_ptr<uint8_t[]>` is incompatible with `std::unique_ptr<UWB_APP_CONFIG_PARAM>`; attempting to force it with `static_cast` will point out that one doesn't derive from the other. 

There's no way to allocate a flex-sized `UWB_APP_CONFIG_PARAM` that can be used with `std::unique_ptr`, because the deleter will always be bound to the contained type. 

There needs to be two stages: one is allocating the byte buffer, you've got that already, the second is interpreting the buffer as the intended type `T`. You could do something like this:

```suggestion
template <class T>
std::tuple<std::unique_ptr<uint8_t[]>, UWB_APP_CONFIG_PARAM*, T*>
GetUWB_APP_CONFIG_PARAM(const T &value)
{
    auto appConfigSize = offsetof(UWB_APP_CONFIG_PARAM, paramValue[sizeof(T)]);
    auto appConfigBuffer = std::make_unique<uint8_t[]>(appConfigSize);
    auto appConfigParam = reinterpret_cast<UWB_APP_CONFIG_PARAM*>(appConfigBuffer.get());
    auto appConfigValue = reinterpret_cast<T&>(&appConfigParam->paramValue[0]);
    return { appConfigBuffer, appConfigParam, appConfigValue };
}
```

This gets messy though, since the ownership of the tuple contents is unclear, and they could get passed around individually and easily lose their meaning/binding. An adapter class could manage all of that. Eg.

```
template <class T>
class UwbAppConfigAdapter
{
    UwbAppConfigAdapter(const T& value) : 
        m_buffer(std::make_unique<uint8_t[]>(offsetof(UWB_APP_CONFIG_PARAM), paramValue[sizeof(T)])),
        m_parameter(*reinterpret_cast<UWB_APP_CONFIG_PARAM*>(m_buffer.get()))
        m_value(*reinterpret_cast<T*>(&m_parameter.paramValue[0]))
    {
        m_value = value; // set value contents
    }
    
    T& Value() { return m_value; }
    UWB_APP_CONFIG_PARAM& Parameter() { m_parameter; }
    
private:
    T& m_value;
    UWB_APP_CONFIG_PARAM &m_parameter;
    std::unique_ptr<uint8_t[]> m_buffer;
};
```
Which then gets used as:

```
UwbAppConfigAdapter<uwb::protocol::fira::DeviceRole> uwbAppConfigDeviceRole { uwb::protocol::fira::DeviceRole::Initiator };
auto role = uwbAppConfigDeviceRole.Value();
auto appConfigParam = uwbAppConfigDeviceRole.Parameter(); // pass me to DeviceIoControl()
```

abeltrano: #23

windows/devices/uwb/UwbSession.cxx
```diff
@@ -32,6 +174,57 @@ UwbSession::ConfigureImpl(const uwb::protocol::fira::UwbSessionData& uwbSessionD
     }
 
     m_sessionId = uwbSessionData.sessionId;
+
+    // Populate the PUWB_SET_APP_CONFIG_PARAMS
+    auto sessionUwbMap = uwbSessionData.uwbConfiguration.GetValueMap();
+
+    std::vector<std::shared_ptr<uint8_t>> datavector(ddi_service_map.size());
```
abeltrano, line 181: I'd like to see this code move to a helper function in some form whose job is to take a `UwbSessionData` object and output a `UWB_SET_APP_CONFIG_PARAMS` structure:

```
UWB_SET_APP_CONFIG_PARAMS
GenerateUwbSetAppConfigParameterDdi(const UwbSessionData& uwbSessionData)
```

So far, we've kept this kind of code outside of class'es interfaces and implemented them in detail namespaces in the source files they are needed to separate their concerns (see [here](https://github.com/microsoft/nearobject-framework/blob/27606804546b25b20606f23a045615b2ffbd1232/windows/devices/uwb/UwbDevice.cxx#L42-L49) for an example. In this case, `UwbSessionData` itself shouldn't need to be aware that the UwbCx DDI exists at all.

We should take that one step further and implement them independently in a helper library of some sort (eg. `libuwbcxadaption` or something). This way, we can independently test that functionality and isolate changes to the DDI to just that library.

windows/devices/uwb/UwbSession.cxx
```diff
@@ -14,12 +16,152 @@ UwbSession::UwbSession(std::weak_ptr<uwb::UwbSessionEventCallbacks> callbacks, w
...
+/**
+ * @brief This function only works for integer types, aka uint8_t, uint16_t, uint32_t, uint64_t
+ *
+ * @param value
+ * @param size
+ * @param dst
+ * @return uint8_t*
+ */
+uint8_t *
+WriteBytes(uint64_t value, std::size_t size, uint8_t *dst)
```
abeltrano, line 41: Could this not be replaced with `std::memcpy` and then there would be no restrictions on the input?

windows/devices/uwb/UwbSession.cxx
abeltrano, line 181: Why `std::shared_ptr` instead of `std::unique_ptr`?

windows/devices/uwb/UwbSession.cxx
```diff
@@ -14,12 +16,152 @@ UwbSession::UwbSession(std::weak_ptr<uwb::UwbSessionEventCallbacks> callbacks, w
...
+    memcpy(dst, value.data(), size);
+    return dst + size;
+}
+
+/**
+ * @brief Maps the param to the length of the value. If the value is 0, means a variable length and special code to handle.
+ * Taken from the FiRa UCI spec Table 29:APP Configuration IDs.
+ * TODO put this somewhere more appropriate.
+ */
+std::unordered_map<UWB_APP_CONFIG_PARAM_TYPE, unsigned int> appConfigParamLengths{
```
abeltrano, line 81: I'm not sure I understand why we need this map. The input type has its own size, so we'll always know what it is and how much data we have to write based on the DDI structure field(s) we're filling in.

windows/devices/uwb/UwbSession.cxx
```diff
@@ -14,12 +16,152 @@ UwbSession::UwbSession(std::weak_ptr<uwb::UwbSessionEventCallbacks> callbacks, w
...
+};
+
+std::unordered_map<
+    UWB_APP_CONFIG_PARAM_TYPE,
+    std::tuple<uwb::protocol::fira::UwbConfiguration::ParameterTag,
+        std::function<uint8_t *(std::any, std::size_t, uint8_t *)>>>
+    ddi_service_map{
+        { UWB_APP_CONFIG_PARAM_TYPE_DEVICE_ROLE,
+            { uwb::protocol::fira::UwbConfiguration::ParameterTag::DeviceRole,
+                [](std::any arg, std::size_t size, uint8_t *dst) {
```
abeltrano, line 137: Why do we have to sacrifice type-safety by using `std::any`? Can't we use `std::visit` to get the appropriate types in the `UwbSessionData` value-map?

windows/devices/uwb/UwbSession.cxx
forwardpointer: I don't know if sizeof(enum) is guaranteed to return the right size so I think it's still a good idea to have this map. Especially for those enums of weird sizes like 3 elements (does the internal representation of such an enum always map to a uint8_t?)

windows/devices/uwb/UwbSession.cxx
forwardpointer: No particular reason. I read up on the differences and will change the pointers accordingly.

windows/devices/uwb/UwbSession.cxx
forwardpointer: The use of std::any was before the std::variant pr was merged into develop. I will switch to using variants

tests/unit/windows/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ target_link_libraries(nearobject-test-windows
         nearobject-service-windows
         notstd-windows
         uwb
-        windevuwb
+        libuwbcxadapter
```
abeltrano, line 18: The dependency on `windevuwb` should be restored here. The reason is that we're testing stuff in that library directly in this executable. Even though `libuwbcxadapter` depends on `windevuwb`, not specifying it adds risks to rely on indirect dependencies in case we ever change things (eg. if we decide to split the tests into separate binaries, we'll hit linker issues since not all the explicit depends are listed).

Also, the `lib` prefix isn't needed; the linker automatically prepends dependency names with `lib` when attempting to link, and it is common practice not to specify the prefix.

```suggestion
        uwbcxadapter
        windevuwb
```

windows/devices/uwb/CMakeLists.txt
```diff
@@ -1,3 +1,46 @@
...
+set(LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE ${CMAKE_CURRENT_LIST_DIR}/include)
+set(LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE_PREFIX ${LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE}/windows/devices/uwb)
+
+target_sources(libuwbcxadapter
+    PRIVATE
+        ${CMAKE_CURRENT_LIST_DIR}/UwbCxAdapter.cxx
+        ${CMAKE_CURRENT_LIST_DIR}/UwbAppConfiguration.cxx
+    PUBLIC
+        ${LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE_PREFIX}/UwbCxAdapter.hxx
+        ${LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE_PREFIX}/UwbAppConfiguration.hxx
```
abeltrano, line 12: Please make sure these are sorted in ascending order for consistency. Same goes for any other CMakeLists.txt files that have added files.

```suggestion
        ${CMAKE_CURRENT_LIST_DIR}/UwbAppConfiguration.cxx
        ${CMAKE_CURRENT_LIST_DIR}/UwbCxAdapter.cxx
    PUBLIC
        ${LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE_PREFIX}/UwbAppConfiguration.hxx
        ${LIBUWBCXADAPTER_DIR_PUBLIC_INCLUDE_PREFIX}/UwbCxAdapter.hxx
```

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+    UwbSetAppConfigurationParametersBuilder builder{ uwbSessionData.sessionId };
+
+    // TODO NUMBER_OF_CONTROLEES
+    // TODO DEVICE_MAC_ADDRESS
+    // TODO DST_MAC_ADDRESS
+    // TODO DEVICE_TYPE
+    const auto sessionUwbMap = uwbSessionData.uwbConfiguration.GetValueMap();
+
+    for (const auto [ptag, value] : sessionUwbMap) {
+        std::visit(overloaded{
```
abeltrano, line 120: Nice! I couldn't get this to work 👍

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+using namespace windows::devices;
+
+template <class... Ts>
+struct overloaded : Ts...
+{
+    using Ts::operator()...;
+};
+// explicit deduction guide (not needed as of C++20)
+template <class... Ts>
+overloaded(Ts...) -> overloaded<Ts...>;
```
abeltrano, line 106: Am guessing we don't need this since we're compiling with C++20? If so, let's remove it.
```suggestion
```

windows/devices/uwb/include/windows/devices/uwb/UwbCxAdapter.hxx
```diff
@@ -0,0 +1,22 @@
...
+
+#include <uwb/UwbMacAddress.hxx>
+#include <uwb/UwbSession.hxx>
+#include <uwb/UwbSessionEventCallbacks.hxx>
+#include <uwb/protocols/fira/UwbConfiguration.hxx>
+
+namespace windows::devices::uwb
+{
+std::unique_ptr<uint8_t[]>
+GenerateUwbSetAppConfigParameterDdi(const ::uwb::protocol::fira::UwbSessionData& uwbSessionData);
```
abeltrano, line 19: Missing function comment block here.

windows/devices/uwb/include/windows/devices/uwb/UwbAppConfiguration.hxx
```diff
@@ -97,12 +100,39 @@ private:
...
+    template <typename PropertyT>
+    void
+    Push(UwbAppConfigurationParameter<PropertyT> param){
+       m_params.push_back(std::move(param.m_buffer));
+       m_paramsLengthSum += param.Size();
+    }
+
+    /**
+     * Writes all the params and returns a buffer that can be reinterpreted as a UWB_SET_APP_CONFIG_PARAMS
+    */
```
abeltrano, line 124: Comment blocks need updating for return values (and tabbing looks off).

lib/uwb/protocols/fira/RangingMethod.cxx
```diff
@@ -0,0 +1,20 @@
+
+#include <tuple>
```
abeltrano, line 2: Is this actually needed? I think this was left in there accidentally and can probably be removed.
```suggestion
```

lib/uwb/protocols/fira/RangingMethod.cxx
```diff
@@ -0,0 +1,20 @@
...
+    if (Method == RangingDirection::SingleSidedTwoWay and ReportMode == MeasurementReportMode::Deferred) {
+        return 1;
+    } else if (Method == RangingDirection::DoubleSidedTwoWay and ReportMode == MeasurementReportMode::Deferred) {
+        return 2;
+    } else if (Method == RangingDirection::SingleSidedTwoWay and ReportMode == MeasurementReportMode::NonDeferred) {
+        return 3;
+    } else if (Method == RangingDirection::DoubleSidedTwoWay and ReportMode == MeasurementReportMode::NonDeferred) {
+        return 4;
+    } else {
+        return 0;
```
abeltrano, line 18: This is invalid, right? It seems like we need a way to signal to the caller this is invalid, otherwise, I think they might use this to serialize an invalid value.

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint8_t(arg), service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](uint16_t arg) {
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint16_t(arg), service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](uint32_t arg) {
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint32_t(arg), service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](::uwb::protocol::fira::Channel arg) {
+                           uint8_t data = 9;
```
abeltrano, line 134: If this is supposed to be the default value, then it should probably be assigned in a `default` case label. However, if nothing matches, I'd think we'd want to flag that somehow, not assign a default value.

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+                               break;
+                           }
+                           }
+                           builder.Push(std::move(UwbAppConfigurationParameter{ data, service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](::uwb::protocol::fira::ConvolutionalCodeConstraintLength arg) {
+                           // TODO
+                       },
+                       [&builder, ptag](::uwb::protocol::fira::DeviceRole arg) {
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint8_t(arg), service_ddi_map.at(ptag) }));
```
abeltrano, line 175: All of the `enum class`'es that match parameter tags have been defined such that their underlying type is `uint8_t`. If there is concern about this possibly mismatching, then you can add a `static_assert` to guarantee it, which is never a bad idea (I might do that someplace else, maybe at the very beginning of this function since it's an assumption this function relies on). So, I think `notstd::to_underlying` should be used here (and other similar instances) instead. And if not, at least a `static_cast<uint8_t>` should be preferred over a C-style cast.

windows/devices/uwb/UwbSession.cxx
```diff
@@ -3,9 +3,12 @@
 #include <wil/result.h>
 
 #include <algorithm>
+#include <magic_enum.hpp>
```
abeltrano, line 6: Can this be removed now?

windows/devices/uwb/CMakeLists.txt
```diff
@@ -23,7 +65,9 @@ target_include_directories(windevuwb
 target_link_libraries(windevuwb
     PRIVATE
         windevutil
+        magic_enum::magic_enum
```
abeltrano, line 68: Can this be removed now?
```suggestion
```

windows/devices/uwb/UwbSession.cxx
abeltrano, line 6: Can this be removed now?
```suggestion
```

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+    { UWB_APP_CONFIG_PARAM_TYPE_UWB_INITIATION_TIME, 4 },
+    { UWB_APP_CONFIG_PARAM_TYPE_HOPPING_MODE, 1 },
+    { UWB_APP_CONFIG_PARAM_TYPE_BLOCK_STRIDE_LENGTH, 1 },
+    { UWB_APP_CONFIG_PARAM_TYPE_RESULT_REPORT_CONFIG, 1 },
+    { UWB_APP_CONFIG_PARAM_TYPE_IN_BAND_TERMINATION_ATTEMPT_COUNT, 1 },
+    { UWB_APP_CONFIG_PARAM_TYPE_SUB_SESSION_ID, 4 },
+    { UWB_APP_CONFIG_PARAM_TYPE_BPRF_PHR_DATA_RATE, 1 },
+    { UWB_APP_CONFIG_PARAM_TYPE_MAX_NUMBER_OF_MEASUREMENTS, 2 },
+    { UWB_APP_CONFIG_PARAM_TYPE_STS_LENGTH, 1 }
+};
```
abeltrano, line 59: It looks like this is no longer used and can be removed now.

```suggestion
```

windows/devices/uwb/UwbSession.cxx
abeltrano: It looks like this is no longer used, so the point is probably moot. In general, since `enum class` entities were being used, and all of them had fixed base-type, their underlying type is always as it is explicitly specified, so can be relied upon for a size calculation if desired.

windows/devices/uwb/include/windows/devices/uwb/UwbAppConfiguration.hxx
```diff
@@ -97,12 +100,40 @@ private:
...
+
+    /**
+     * Push a UwbAppConfigurationParameter into the collection
+     */
+    template <typename PropertyT>
+    void
+    Push(UwbAppConfigurationParameter<PropertyT> param)
+    {
+        m_params.push_back(std::move(param.m_buffer));
+        m_paramsLengthSum += param.Size();
```
abeltrano, line 120: This is modifying the input `UwbAppConfigurationParameter` which might be unexpected to the caller since it will leave that object in an invalid state (no buffer, but `Size()` returns the size of the prior buffer contents). 

In the end in `Publish` all of this data is copied anyway, so might it be simpler to just maintain one large buffer and copy the data into it on each `Push`? For example, instead of `std::vector<std::unique_ptr<uint8_t[]>> m_params;` we instead use `std::vector<uint8_t> m_params;` and then `Push` gets implemented like this:

```suggestion
        m_params.insert(std::end(m_params), param.Buffer(), param.Buffer() + param.Size());
```

This way you don't need the `m_paramsLengthSum` either, since `m_params.size()` tracks that.

windows/devices/uwb/include/windows/devices/uwb/UwbAppConfiguration.hxx
```diff
@@ -7,26 +7,29 @@
...
- * 
+ * @brief Adaptor for using UWB_APP_CONFIG_PARAM UwbCx DDI structure in a nicer way.
+ *
  * @tparam PropertyT The type of property the UWB_APP_CONFIG_PARAM structure
  * holds as its flexible array member.
  */
 template <typename PropertyT>
 class UwbAppConfigurationParameter
 {
+    friend class UwbSetAppConfigurationParametersBuilder; // friend class so the unique buffer ptr can be moved out
```
abeltrano, line 26: See my later comment about not `move`'ing this out. If you agree with that, I think this can be removed.

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+
+const std::unordered_map<
+    ::uwb::protocol::fira::UwbConfiguration::ParameterTag,
+    UWB_APP_CONFIG_PARAM_TYPE>
+    service_ddi_map{
```
abeltrano, line 64: This map always gets used as an argument to the `UwbAppConfigurationParameter` constructor. Might it make sense to have the `UwbAppConfigurationParameter` constructor use the map directly and simplify its interface to take the parameter tag instead? Eg. Change this:

```
explicit UwbAppConfigurationParameter(const PropertyT& value, UWB_APP_CONFIG_PARAM_TYPE parameterType, size_t parameterSize = sizeof(PropertyT)) 
```

to this:
```
explicit UwbAppConfigurationParameter(const PropertyT& value, UwbConfiguration::ParameterTag parameterTag, size_t parameterSize = sizeof(PropertyT)) 
```

Then the calling code can do this:
```
builder.Push(UwbAppConfigurationParameter{ static_cast<uint8_t>(arg), ptag }));
```

The caller is then required to know absolutely nothing about the UwbCx details like `UWB_APP_CONFIG_PARAM_TYPE `, which is especially nice since this adapter is supposed to help working with the UwbCx.

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+                           }
+                           case (::uwb::protocol::fira::Channel::C13): {
+                               data = 13;
+                               break;
+                           }
+                           case (::uwb::protocol::fira::Channel::C14): {
+                               data = 14;
+                               break;
+                           }
+                           }
```
abeltrano, line 168: If the value always maps to the logical channel number, then let's update the `enum class Channel` to have `uint8_t` as its base/underlying type, and explicitly set the corresponding values to what the spec defines, eg.

```
enum class Channel : uint8_t {
    C5 = 5,
    C6 = 6,
    C8 = 8,
    C9 = 9,
    C10 = 10,
    C12 = 12,
    C13 = 13,
    C14 = 14,
};
```

Then all of this code can be condensed to:
```suggestion
                           uint8_t data = notstd::to_underlying(arg);
```

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+                               }
+                               case ::uwb::protocol::fira::ResultReportConfiguration::AoAElevationReport: {
+                                   result |= 0x4;
+                                   break;
+                               }
+                               default: {
+                                   result |= 0x8;
+                                   break;
+                               }
+                               }
```
abeltrano, line 231: Instead of having these magic numbers, let's update `ResultReportConfiguration` and use the underlying value per below:
```
enum class ResultReportConfiguration : uint8_t {
    TofReport = 0x01U,
    AoAAzimuthReport = 0x02U,
    AoAElevationReport  0x04U,
    AoAFoMReport = 0x08U,
};
```

Then used like this:
```suggestion
                               result |= notstd::to_underlying(config);
```

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,239 @@
...
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint8_t(arg), service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](::uwb::protocol::fira::StsConfiguration arg) {
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint8_t(arg), service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](::uwb::protocol::fira::StsPacketConfiguration arg) {
+                           builder.Push(std::move(UwbAppConfigurationParameter{ uint8_t(arg), service_ddi_map.at(ptag) }));
+                       },
+                       [&builder, ptag](::uwb::UwbMacAddress arg) {
+                           builder.Push(std::move(UwbAppConfigurationParameter{ arg.GetValue(), service_ddi_map.at(ptag) }));
```
abeltrano, line 199: This won't work here since `UwbAppConfigurationParameter::PropertyT` will be `std::span<const uint8_t>` and the internal representation of the span will be copied to the value instead of the actual data the span refers to (which is what we want).

We'll need to limit the types `PropertyT` to probably satisfy `std::is_standard_layout`, but that can be done later. For now, you need to pass something that directly represents the data in memory sequentially, like `std::array<uint8_t, N>`, then pass its buffer pointer as the value, and specify the size directly.

windows/devices/uwb/UwbCxAdapter.cxx
forwardpointer: Currently the switch statement does cover all the enum cases. Maybe I'll just throw an error if nothing matches (shouldn't happen)

abeltrano: This becomes moot if we switch to use `notstd::to_underlying`.

windows/devices/uwb/include/windows/devices/uwb/UwbAppConfiguration.hxx
forwardpointer: Oh that's a good idea

windows/devices/uwb/UwbCxAdapter.cxx
forwardpointer: I agree with this

windows/devices/uwb/UwbAppConfiguration.cxx
```diff
@@ -0,0 +1,80 @@
...
+        return totalSize + uwbAppConfigurationParameter->DdiSize();
+    });
+}
+} // namespace detail
+
+UwbSetAppConfigurationParameters::UwbSetAppConfigurationParameters(const std::vector<std::shared_ptr<IUwbAppConfigurationParameter>>& parameters) :
+    m_buffer(offsetof(UWB_APP_CONFIG_PARAMS, appConfigParamsCount) + ::detail::CalculateTotalUwbAppConfigurationBufferSize(parameters)), // TOOD: verify this is correct
+    m_parameters(*reinterpret_cast<UWB_APP_CONFIG_PARAMS*>(std::data(m_buffer)))
+{
+    auto it = std::end(m_buffer);
```
abeltrano, line 61: This is a bug. The data should start to be inserted at `offsetof(UWB_APP_CONFIG_PARAMS, appConfigParamsCount)`.

lib/uwb/protocols/fira/RangingMethod.cxx
forwardpointer: I agree. Technically the values 0x0 and 0x5 thru 0xFF are Reserved for Future Use so we should throw an error or something in case in the future those values get used for something else.

lib/uwb/protocols/fira/RangingMethod.cxx
```diff
@@ -0,0 +1,18 @@
...
+    if (Method == RangingDirection::SingleSidedTwoWay and ReportMode == MeasurementReportMode::Deferred) {
+        return 1;
+    } else if (Method == RangingDirection::DoubleSidedTwoWay and ReportMode == MeasurementReportMode::Deferred) {
+        return 2;
+    } else if (Method == RangingDirection::SingleSidedTwoWay and ReportMode == MeasurementReportMode::NonDeferred) {
+        return 3;
+    } else if (Method == RangingDirection::DoubleSidedTwoWay and ReportMode == MeasurementReportMode::NonDeferred) {
+        return 4;
+    } else {
+        throw std::runtime_error::runtime_error("invalid ranging method");
```
abeltrano, line 16: The second `runtime_error` isn't needed:
```suggestion
        throw std::runtime_error("invalid ranging method");
```

windows/devices/uwb/UwbCxAdapter.cxx
```diff
@@ -0,0 +1,64 @@
...
+    // TODO DEVICE_MAC_ADDRESS
+    // TODO DST_MAC_ADDRESS
+    // TODO DEVICE_TYPE
+    const auto sessionUwbMap = uwbSessionData.uwbConfiguration.GetValueMap();
+
+    for (const auto& [parameterTag, parameterValueVariant] : sessionUwbMap) {
+        std::visit([&](auto&& parameterValue) {
+            using ParameterValueT = std::decay_t<decltype(parameterValue)>;
+            if constexpr (std::is_enum_v<ParameterValueT>) {
+                parameters.push_back(std::make_shared<UwbAppConfigurationParameter<std::underlying_type_t<ParameterValueT>>>(notstd::to_underlying(parameterValue), parameterTag));
```
forwardpointer, line 25: Is there a reason why this is shared_ptr?

windows/devices/uwb/UwbAppConfiguration.cxx
```diff
@@ -0,0 +1,92 @@
...
+    m_buffer(offsetof(UWB_APP_CONFIG_PARAMS, appConfigParams[0]) + ::detail::CalculateTotalUwbAppConfigurationBufferSize(parameters)), // TOOD: verify this is correct
+    m_parameters(*reinterpret_cast<UWB_SET_APP_CONFIG_PARAMS*>(std::data(m_buffer)))
+{
+    auto it = std::next(std::begin(m_buffer), offsetof(UWB_APP_CONFIG_PARAMS, appConfigParams[0]));
+
+    // Copy each complete UWB_APP_CONFIG_PARAM structure into the byte buffer.
+    for (const auto& parameter : parameters) {
+        const auto& parameterBuffer = parameter->DdiBuffer();
+        m_buffer.insert(it, std::cbegin(parameterBuffer), std::cend(parameterBuffer));
+        std::advance(it, std::size(parameterBuffer));
```
forwardpointer, line 74: I think we should use a simple byte array for m_buffer instead of a vector. Here, we already know exactly which parameters go in this struct, so there is no need for a dynamic vector. In addition, m_buffer.insert invalidates all iterators if more memory has been allocated due to its operation.

windows/devices/uwb/UwbCxAdapter.cxx
abeltrano: Yes. `UwbAppConfigurationParameter` is a template class, which means it's not a real class; each `PropertyT` type it's instantiated with creates a unique type. Therefore, we cannot store ``UwbAppConfigurationParameter` objects in a container. So to alleviate that, we introduced a _non-templated_ base class `IUwbAppConfigurationParameter` that `UwbAppConfigurationParameter` derives from. We can store instances of `IUwbAppConfigurationParameter` in a container.

However, if we stored the base class directly in the container, then when we attempt to access the actual derived implementation, it'll fail due to object slicing. However, polymorphic access works through base class _pointers_., so we allocated and access the objects dynamically via pointer instead.

Any time you're doing dynamic allocation, one of the smart pointer types should be used to manage the allocated memory safely: `unique_ptr` if used in isolation, otherwise `shared_ptr` if it's passed around. In this case, the object is passed around, so `shared_ptr` was used.

windows/devices/uwb/UwbAppConfiguration.cxx
abeltrano: The use of `std::vector` has many benefits over a simple byte array: use with other STL algorithms, re-sizing if necessary, easy copying to other containers, the list goes on.

Just because we don't need the grow the buffer in this instance, doesn't mean we shouldn't make use of the other features `std:;vector` has to offer. I might argue the use of `std::vector` here is worth it just for the fact that it contains both the data and the size together in one object. With a simple byte-buffer, we don't have that, and the state (data contents, size) could get out of sync.

Regarding `insert`: the vector was constructed with the exact number of bytes needed, so no re-allocation or iterator invalidations occur here.

tests/unit/windows/TestUwbAppConfiguration.cxx
```diff
@@ -114,8 +133,67 @@ TEST_CASE("UwbAppConfiguration performs allocation for contained value correctly
...
 
-        auto& appConfigurationDdi = appConfiguration.DdiParameter();
-        auto& dstMacAddressActual = appConfiguration.Value();
+        const auto& appConfigurationDdi = appConfiguration.DdiParameter();
+        const auto& dstMacAddressActual = appConfiguration.Value();
         REQUIRE(dstMacAddressActual == dstMacAddressExpected);
     }
 }
+
+TEST_CASE("UwbSetAppConfigurationParameters performs allocation for contained values correctly", "[basic]")
```
abeltrano, line 142: You might already be working on it, but we'll definitely want an explicit test case for each of the unique types being converted: enum, integral (if any), `RangingMethod`, `std::unordered_set<::uwb::protocol::fira::ResultReportConfiguration>>`, `UwbMacAddress` with both lengths (short [2], extended [8]). Those types come from the `std::visit` function in `GenerateUwbSetAppConfigParameterDdi`.

windows/devices/uwb/UwbAppConfiguration.cxx
forwardpointer: I'm fairly certain that insert always changes the size of vector. When I was debugging that was certainly the case. If we really want to use vector then we can't use insert here, but instead a for-each statement or something

forwardpointer: Also, the byte array is intended to hold a structure that does have an internal size field, so even though it's not pretty I think the benefit of the vector object also holding the size is not that big.

windows/devices/uwb/UwbCxAdapter.cxx
forwardpointer: Ah I see, the interface is so that we don't have to use the Builder model. I guess I'm having difficulty differentiating use cases where objects are owned by several different things vs not. In this case, it appears that first the parameters vector owns the pointers, but then all the pointers are passed into the constructor of UwbSetAppConfigurationParameters. Thus it appears to me that at any point in time they are only owned by one thing.

windows/devices/uwb/UwbAppConfiguration.cxx
abeltrano: > I'm fairly certain that insert always changes the size of vector. When I was debugging that was certainly the case. If we really want to use vector then we can't use insert here, but instead a for-each statement or something

Sorry, you're absolutely right about this. I mixed up `vector::insert` and `std::inserter`, the latter of which creates an `std::insert_iterator` which is used to insert at the specified iterator position. As you mentioned, the fix for the insertion is a range-based for works or `std::transform`.

> Also, the byte array is intended to hold a structure that does have an internal size field, so even though it's not pretty I think the benefit of the vector object also holding the size is not that big. Regarding the out-of-sync of the (data contents, size) state, that only happens right when it is initialized, and never again afterwards (unless the user deliberately overwrites the size field of the byte array).

These are all good points. Even so, is there any reason not to use `std::vector`; the benefit is clearly non-zero, so why not take that as win and go with it? `std::vector` is better then `std::unique_ptr<uint8_t[]>` in every case and scenario I can imagine, and I can't think of any place where `unique_ptr` would be preferred.

windows/devices/uwb/UwbCxAdapter.cxx
abeltrano: Actually yes, I think you're right about this, so probably should be `std::unique_ptr` here instead. I had been toying with something else where these were being passed around, but that's certainly not the case right now.

windows/devices/uwb/UwbAppConfiguration.cxx
forwardpointer: I agree with the statement about no real drawback to using vector, and I switched back to using vector

windows/devices/uwb/UwbCxAdapter.cxx
forwardpointer: I switched to using unique_ptr

tests/unit/uwb/protocols/fira/TestUwbFiraUwbConfigurationBuilder.cxx
```diff
@@ -15,7 +15,7 @@ constexpr uint32_t FiraVersionMacValue{ 0x1234ABCD };
 constexpr uint32_t FiraVersionPhyValue{ 0xDCBA4321 };
 constexpr Channel ChannelValue{ uwb::protocol::fira::Channel::C10 };
 constexpr DeviceRole DeviceRoleValue{ DeviceRole::Initiator };
-constexpr RangingMethod RangingMethodValue{ RangingDirection::OneWay, MeasurementReportMode::Deferred };
+constexpr RangingMethod RangingMethodValue{ RangingDirection::SingleSidedTwoWay, MeasurementReportMode::Deferred };
```
abeltrano, line 18: Just curious: why this change?

tests/unit/windows/TestUwbAppConfiguration.cxx
```diff
@@ -114,8 +135,316 @@ TEST_CASE("UwbAppConfiguration performs allocation for contained value correctly
...
+
+        REQUIRE(ddiParams.sessionId == sessionData.sessionId);
+        REQUIRE(ddiParams.appConfigParamsCount == 1);
+
+        const auto param = ddiParams.appConfigParams[0];
+        const auto sessionUwbMap = sessionData.uwbConfiguration.GetValueMap();
+
+        std::vector<std::unique_ptr<IUwbAppConfigurationParameter>> parameters;
+
+        for (const auto& [parameterTag, parameterValueVariant] : sessionUwbMap) {
```
abeltrano, line 269: Like the `enum` test case above, let's get rid of this loop since it's known there's only 1 value. You can use the value helper accessor function `GetControleeShortMacAddress()` to get the value instead.

tests/unit/windows/TestUwbAppConfiguration.cxx
```diff
@@ -114,8 +135,316 @@ TEST_CASE("UwbAppConfiguration performs allocation for contained value correctly
...
+            const auto& ddiParams = testStruct.DdiParameters();
+
+            REQUIRE(ddiParams.sessionId == sessionData.sessionId);
+            REQUIRE(ddiParams.appConfigParamsCount == std::size(parameters));
+            REQUIRE(ddiParams.size == parameters[0]->DdiSize() + offsetof(UWB_SET_APP_CONFIG_PARAMS, appConfigParams[0]));
+            REQUIRE(std::memcmp(ddiParams.appConfigParams, parameters[0]->DdiBuffer().data(), parameters[0]->DdiSize()) == 0);
+        }
+    }
+
+    SECTION("mac address works")
```
abeltrano, line 247: This is good for a short mac.

We also need to test an extended mac. Can use the `UwbConfiguration::Builder::SetMacAddressController()` function to do that, and the parameter tag is `UwbConfiguration::ParameterTag::ControllerMacAddress`.

tests/unit/windows/TestUwbAppConfiguration.cxx
```diff
@@ -114,8 +135,316 @@ TEST_CASE("UwbAppConfiguration performs allocation for contained value correctly
...
+            const auto& ddiParams = testStruct.DdiParameters();
+
+            REQUIRE(ddiParams.sessionId == sessionData.sessionId);
+            REQUIRE(ddiParams.appConfigParamsCount == std::size(parameters));
+            REQUIRE(ddiParams.size == parameters[0]->DdiSize() + offsetof(UWB_SET_APP_CONFIG_PARAMS, appConfigParams[0]));
+            REQUIRE(std::memcmp(ddiParams.appConfigParams, parameters[0]->DdiBuffer().data(), parameters[0]->DdiSize()) == 0);
+        }
+    }
+
+    // TODO UWB_APP_CONFIG_PARAM_TYPE_RANGING_ROUND_CONTROL
```
abeltrano, line 364: Is this still outstanding? If not, can it be removed?

abeltrano (changes requested):

tests/unit/uwb/protocols/fira/TestUwbFiraUwbConfigurationBuilder.cxx
forwardpointer: I believe OneWay, Deferred is not a valid RangingMethod so I changed this test to use a valid RangingMethod.

tests/unit/windows/TestUwbAppConfiguration.cxx
forwardpointer: Still outstanding. I'm not sure the role of this parameter. It seems to be related to RangingMethod but I'm not sure

tests/unit/windows/TestUwbAppConfiguration.cxx
```diff
@@ -114,8 +135,151 @@ TEST_CASE("UwbAppConfiguration performs allocation for contained value correctly
...
+
+        UwbSessionData sessionData;
+        sessionData.sessionId = 1;
+        sessionData.uwbConfiguration = UwbConfiguration::Create().DeviceRole(roleExpected);
+        sessionData.uwbConfigurationAvailable = true;
+
+        auto params = GenerateUwbSetAppConfigParameterDdi(sessionData);
+        const auto& ddiParams = params.DdiParameters();
+
+        REQUIRE(ddiParams.sessionId == sessionData.sessionId);
```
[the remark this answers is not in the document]
forwardpointer: Agreed and I removed all instances of this check

forwardpointer:
There appears to be a bug uncovered by the tests, will investigate right away.

abeltrano (dismissed):

abeltrano:
> There appears to be a bug uncovered by the tests, will investigate right away.

Awesome. Let's create a new issue for the bug, merge this PR, then investigate separately.

forwardpointer:
The bug is that GenerateUwbSetAppConfigParameterDdi for some reason is using uint8_t for the numeric parameters of type uint16_t and uint32_t

forwardpointer:
I know the reason for the bug but after I switched the UwbAppConfigurationBuilder to use the right integer sizes for RangingInterval and SlotDuration the program uwb-test.exe is aborting, and idk why

abeltrano (approved):

Issue #57, opened 2023-01-31
Bug: Fix plog to use explicit version


abeltrano:
**Describe the bug**
When an explicit plog version is specified, the driver build fails to find v1.1.9 and is stuck on pulling down v1.1.8.

**Steps To Reproduce**
Build the driver.

**Expected behavior**
Driver builds successfully.

**Actual behavior**
Driver fails to build, indicating it cannot find version 1.1.8. The CMake config file lists the version as 1.1.8.

PR #58, opened 2023-01-31
Implement uwb simulator driver dispatch handling for the LRP DDI

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Allow IOCTLs to be properly dispatched to callbacks for the LRP DDI.

### Technical Details

* Move DDI conversions functions to the adaption library.
* Implement dispatching to callbacks.
* Change `plog` to be version-less. This is to address a bug where the driver project is stuck on obtaining v1.1.8 even though v1.1.9 works fine (locally in CMake projects, and in Windows). This issue is being tracked with #57.
* Rename UwbCx LRP DDI "glue" header into something more reflective of its purpose.
* Complete driver requests when there is a failure to obtain the i/o buffers.

### Test Results

* Ad-hoc testing with the driver so far, more TBD.

### Reviewer Focus

None

### Checklist


windows/devices/uwb/UwbCxAdapterDdiLrp.cxx
```diff
@@ -0,0 +1,106 @@
...
+    std::visit([&status](auto&& arg) {
+        using T = std::decay_t<decltype(arg)>;
+        if constexpr (std::is_same_v<T, UwbStatusGeneric>) {
+            status = StatusMapGeneric.at(arg);
+        } else if constexpr (std::is_same_v<T, UwbStatusSession>) {
+            status = StatusMapSession.at(arg);
+        } else if constexpr (std::is_same_v<T, UwbStatusRanging>) {
+            status = StatusMapRanging.at(arg);
+        } else {
+            throw std::runtime_error("unknown UwbStatus variant value encountered");
```
forwardpointer, line 62: Maybe we could log here as well
```PLOG_ERROR << "unknown UwbStatus variant value encountered";```

Comment (forwardpointer) [reply windows/devices/uwb/UwbCxAdapterDdiLrp.cxx:62]: If so, we'd have to include the plog header
```#include <plog/Log.h>```

File: windows/devices/uwb/UwbCxAdapterDdiLrp.cxx [hunk 2]
```diff
@@ -0,0 +1,106 @@
...
+        return UWB_SESSION_STATE_INIT;
+    case UwbSessionState::Deinitialized:
+        return UWB_SESSION_STATE_DEINIT;
+    case UwbSessionState::Active:
+        return UWB_SESSION_STATE_ACTIVE;
+    case UwbSessionState::Idle:
+        return UWB_SESSION_STATE_IDLE;
+    }
+
+    throw std::runtime_error("unknown enumeration value");
```
forwardpointer, line 84: log here?

forwardpointer (approved):

windows/devices/uwb/UwbCxAdapterDdiLrp.cxx
abeltrano: For these cases, it would be a developer bug if we hit them, so the intention is to crash the program immediately so it can't be missed instead of logging. Same with prior suggestion of replacing `std::runtime_error()` with log.

PR #71, opened 2023-02-08
Plog Wrapper for ETW trace logging

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Implement a wrapper for etw trace logging provider.
Use the provider in NearObjectSvc
Rename the logging utils library (which at the moment only provides the log file name generation function).

### Reviewer Focus

Each instance of the TraceLoggingAppender will try to register the global handle again. This could pose a problem if multiple libraries try to use the appender.

The wrapper has a static guid. This means any two services that use this wrapper for etw support will have the same provider guid. The two providers can potentially be differentiated either through plog's auto-logging of the function and LOC, or through additional information the drivers provide themselves in each message.

### Checklist


windows/shared/logging/PlogTraceLogging/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+)
+
+target_include_directories(plog-trace-logging
+    PUBLIC
+        ${PLOGTRACELOGGING_DIR_PUBLIC_INCLUDE}
+)
+
+target_link_libraries(plog-trace-logging
+PUBLIC
+    plog::plog
```
abeltrano, line 21: Looks like tabbing is off here.
```suggestion
    PUBLIC
        plog::plog

```

windows/shared/logging/PlogTraceLogging/PlogTraceLogging.cxx
```diff
@@ -0,0 +1,13 @@
+
+#include <chrono>
+#include <iomanip>
+#include <sstream>
+
+#include <PlogTraceLogging/PlogTraceLogging.hxx>
```
abeltrano, line 6: None of these should be needed here, can they be removed? Also, the header that defines `TRACELOGGING_DEFINE_PROVIDER` (`TraceLoggingProvider.h`?) seems to be missing, so that should be added.

abeltrano, line 6: None of these should be needed here, can they be removed? The only `#include`s in a header should be those that define types used in the header. Sometimes for legacy reasons, an exception to this guideline is made for the windows types which are declared in `windows.h` but the order of inclusion of this header can affect other code, so is required to be explicitly included by the caller instead.

In this case, this header doesn't define anything which uses types from `chrono`, `iomanip`, nor `sstream`.

Also, the header that defines `TRACELOGGING_DEFINE_PROVIDER` (`TraceLoggingProvider.h`?) seems to be missing, so that should be added. This also requires `windows.h`, so that needs to be added as well.

```suggestion
#include <windows.h>

#include <TraceLoggingProvider.h>
```

windows/shared/logging/PlogTraceLogging/include/PlogTraceLogging/PlogTraceLogging.hxx
```diff
@@ -0,0 +1,46 @@
...
+#include <plog/Log.h>
+
+#include <string>
+
+TRACELOGGING_DECLARE_PROVIDER(g_hplogTraceLoggingProvider);
+
+namespace plog
+{
+
+template <class Formatter>
```
abeltrano, line 19: The rest of the codebase is using `typename` instead of `class`, so let's update this to be consistent:
```suggestion
template <typename Formatter>
```

`typename` is more accurate than `class` which is why we prefer it in the codebase (eg. not all template parameter values are classes, but they all are type names).

windows/shared/logging/PlogTraceLogging/include/PlogTraceLogging/PlogTraceLogging.hxx
```diff
@@ -0,0 +1,46 @@
...
+    TraceLoggingAppender()
+    {
+        TraceLoggingRegister(g_hplogTraceLoggingProvider);
+    }
+
+    ~TraceLoggingAppender()
+    {
+        TraceLoggingUnregister(g_hplogTraceLoggingProvider);
+    }
+
```
abeltrano, line 32: Let's add a comment block here describing how this is essentially a relay to a configured Windows TraceLoggingProvider.

windows/shared/logging/PlogTraceLogging/include/PlogTraceLogging/PlogTraceLogging.hxx
```diff
@@ -0,0 +1,46 @@
...
+public:
+    TraceLoggingAppender()
+    {
+        TraceLoggingRegister(g_hplogTraceLoggingProvider);
+    }
+
+    ~TraceLoggingAppender()
+    {
+        TraceLoggingUnregister(g_hplogTraceLoggingProvider);
+    }
```
abeltrano, line 31: This makes the appender a bit inflexible. Why not introduce a member variable which is populated with the ctor, then use that for thr `TraceLoggingRegister/Unregister` calls instead? This way, the appender isn't dependent on a global, and can be configured per-client as needed.

```suggestion
public:
    TraceLoggingAppender(const GUID& traceLoggingProvider) :
        m_traceLoggingProvider(traceLoggingProvider)
    {
        TraceLoggingRegister(m_traceLoggingProvider);
    }

    ~TraceLoggingAppender()
    {
        TraceLoggingUnregister(m_traceLoggingProvider);
    }
    
private:
    GUID m_traceLoggingProvider;
```

windows/shared/logging/PlogTraceLogging/include/PlogTraceLogging/PlogTraceLogging.hxx
```diff
@@ -0,0 +1,46 @@
...
+namespace plog
+{
+
+template <class Formatter>
+class TraceLoggingAppender : public IAppender
+{
+public:
+    TraceLoggingAppender()
+    {
+        TraceLoggingRegister(g_hplogTraceLoggingProvider);
```
abeltrano, line 25: The result of this should probably be checked. Per the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/traceloggingprovider/nf-traceloggingprovider-traceloggingregister), you can use the `SUCCEEDED()` macro on the result.

abeltrano (changes requested):

windows/shared/logging/PlogTraceLogging/PlogTraceLogging.cxx
```diff
@@ -0,0 +1,13 @@
...
+
+
+TRACELOGGING_DEFINE_PROVIDER(
+    g_hplogTraceLoggingProvider,
+    "PlogTraceLoggingProvider",
```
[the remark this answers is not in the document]
abeltrano: See other comments about potentially eliminating this global completely.

windows/shared/logging/PlogTraceLogging/PlogTraceLogging.cxx
forwardpointer: The PlogTraceLogging.hxx header file provides the windows.h and TraceLoggingProvider.h headers

windows/shared/logging/PlogTraceLogging/include/PlogTraceLogging/PlogTraceLogging.hxx
forwardpointer: I don't think TraceLoggingRegister takes a simple GUID, but a TraceLoggingHProvider which would require users of the wrapper to separately define the handle and provide it to this Appender. I think that might make the wrapper of TraceLogging.h kind of pointless then.

forwardpointer: I switched the code to accept a TraceLoggingHProvider, which enables users the flexibility of defining their own guid so we can have different providers for each main.cxx that wants to use etw logging.

windows/shared/logging/PlogTraceLogging/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+    ${PLOGTRACELOGGING_DIR_PUBLIC_INCLUDE_PREFIX}/PlogTraceLogging.hxx
+)
+
+set_target_properties(plog-trace-logging PROPERTIES FOLDER shared/PlogTraceLogging)
+set_target_properties(plog-trace-logging PROPERTIES PUBLIC_HEADER "${PLOGTRACELOGGING_PUBLIC_HEADERS}")
+
+install(
+    TARGETS plog-trace-logging
+    EXPORT plog-trace-logging
+    ARCHIVE
```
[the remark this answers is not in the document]
forwardpointer: When I tried compiling without those three lines I got an error saying the install function doesn't recognize sub-command PUBLIC_HEADERS

windows/shared/logging/PlogTraceLogging/include/PlogTraceLogging/PlogTraceLogging.hxx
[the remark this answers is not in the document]
forwardpointer: I don't think it's appropriate for it to be in a source file because it's part of a templated class, which by nature doesn't generate any source code until it is instantiated with a template parameter

lib/shared/logging/LogUtils/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+)
+
+set_target_properties(LogUtils PROPERTIES FOLDER shared/LogUtils)
+set_target_properties(LogUtils PROPERTIES PUBLIC_HEADER "${LOGUTILS_PUBLIC_HEADERS}")
+
+install(
+    TARGETS LogUtils
+    EXPORT LogUtils
+    ARCHIVE
+    PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/LogUtils
```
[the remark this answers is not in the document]
forwardpointer: I made the change so now the headers can be included with
```
#include <logging/LogUtils.hxx>
#include <logging/windows/PlogTraceLogging.hxx>
```

abeltrano (approved):

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogTraceLogging.hxx
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param providerHandle
+     */
+    TraceLoggingAppender(TraceLoggingHProvider providerHandle) :
+        m_providerHandle(providerHandle)
+    {
+        HRESULT hr = TraceLoggingRegister(m_providerHandle);
+        if (!SUCCEEDED(hr)) {
+            // todo figure out what to do
+            throw std::runtime_error("could not register TraceLoggingProvider!");
```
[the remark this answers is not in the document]
forwardpointer: plog doesn't seem to have error handling setup so I just printed to std::cerr

PR #72, opened 2023-02-09
Implement wrapper for plog to output stuff to the debugger

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

- Implement wrapper for plog to output stuff to the debugger
- Put the debug wrapper functionality in logging-plog-windows library
- make logutils a dependency of logging-plog-windows

### Technical Details

Decided to make the wrapper unconditionally add the DebugAppender.

### Checklist


windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
```diff
@@ -0,0 +1,45 @@
...
+
+#include <string>
+
+namespace logging
+{
+namespace plog
+{
+
+template <typename Formatter>
+class DebugWrapperAppender : public ::plog::DynamicAppender
```
abeltrano, line 19: This is a cool idea 👍

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
```diff
@@ -0,0 +1,45 @@
...
+     * @param providerHandle
+     */
+    DebugWrapperAppender(::plog::IAppender *appender)
+    {
+        ::plog::DynamicAppender::addAppender(appender);
+#ifdef DEBUG
+        if (IsDebuggerPresent()) {
+            ::plog::DynamicAppender::addAppender(&m_debugAppender);
+        }
+#endif
```
abeltrano, line 35: Note that a debugger can be attached to a non-debug build where `DEBUG` isn't defined. Given that, it doesn't seem this wrapper is necessary; we can just add a `::plog::DebugOutputAppender` to the chain of appenders all the time (on Windows). I looked at its implementation and all it does is call `OutputDebugStringW` which is a no-op when no debugger is attached. So, we don't even need to check `IsDebuggerPresent()` like I thought.

windows/tools/devicemon/Main.cxx
```diff
@@ -63,10 +68,10 @@ main(int argc, char* argv[])
...
     }
 
-    windows::devices::DevicePresenceMonitor monitor{deviceGuid.value(), [](auto&& presenceEvent, auto&& deviceName) {
-        const auto presenceEventName = magic_enum::enum_name(presenceEvent);
-        std::cout << deviceName << " " << presenceEventName << std::endl;
-    }};
+    windows::devices::DevicePresenceMonitor monitor{ deviceGuid.value(), [](auto&& presenceEvent, auto&& deviceName) {
+                                                        const auto presenceEventName = magic_enum::enum_name(presenceEvent);
+                                                        std::cout << deviceName << " " << presenceEventName << std::endl;
+                                                    } };
```
abeltrano, line 74: Looks like tabbing is off here.

abeltrano (changes requested):

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
forwardpointer: I decided that it would be better to have this simple wrapper rather than repeat the same 4 lines of constructing the DebugAppender, the DynamicAppender, and adding the DebugAppender and the normal appender to the DynamicAppender. It would also make it easier in the future to extend the wrapper to linux if we wanted to.

PR #77, opened 2023-02-14
Plog debug wrapper

forwardpointer:
This is the same code as PR #72. I just forgot to update the destination branch to develop instead of plog_etw, so this will need to be merged again.


windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
```diff
@@ -0,0 +1,41 @@
...
+     * @brief Construct a new DebugWrapperAppender object
+     * This class adds a DebugOutputApender iff the DEBUG macro is set and IsDebuggerPresent() is true
+     *  otherwise this class just does the same as the input Appender
+     * @param providerHandle
+     */
+    DebugWrapperAppender(::plog::IAppender *appender)
+    {
+        ::plog::DynamicAppender::addAppender(appender);
+        ::plog::DynamicAppender::addAppender(&m_debugAppender);
+    }
```
abeltrano, line 32: Every logger implements `IAppender`, so do we need to take in an `IAppender` as input? For example, could we add this appender to the main logger that's configured with the `RollingFileAppender`? Eg.

```
using LogFormatterType = plog::TxtFormatter;
plog::RollingFileAppender<LogFormatterType> rollingFileAppender(logging::GetLogName("devicemon").c_str());
plog::DebugWrapperAppender<LogFormatterType> debugAppender;
plog::init(plog::verbose, &rollingFileAppender).addAppender(debugAppender); 
```

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
```diff
@@ -0,0 +1,41 @@
...
+
+template <typename Formatter>
+class DebugWrapperAppender : public ::plog::DynamicAppender
+{
+public:
+    /**
+     * @brief Construct a new DebugWrapperAppender object
+     * This class adds a DebugOutputApender iff the DEBUG macro is set and IsDebuggerPresent() is true
+     *  otherwise this class just does the same as the input Appender
+     * @param providerHandle
```
abeltrano, line 26: Looks like the c'tor comment needs updating. Also, we should probably move the comment about the general behavior of this class to the class declaration itself:

```suggestion
/**
 * Common class which routes log messages to an attached debugger for any operating system. 
 * Currently, only Windows is supported.
 */
template <typename Formatter>
class DebugWrapperAppender : public ::plog::DynamicAppender
{
public:
    /**
     * @brief Construct a new DebugWrapperAppender object
     * @param providerHandle
```

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
```diff
@@ -0,0 +1,41 @@
...
+
+#include <string>
+
+namespace logging
+{
+namespace plog
+{
+
+template <typename Formatter>
+class DebugWrapperAppender : public ::plog::DynamicAppender
```
abeltrano, line 19: Why does this need to be a `DynamicAppender` instead of `IAppender`? My understanding is that `DynamicAppender` is needed for adding an appender safely at runtime after initialization, but I'm not sure we have any use case where that's needed. Do you have one in mind?

windows/tools/devicemon/Main.cxx
```diff
@@ -63,10 +68,12 @@ main(int argc, char* argv[])
...
-    windows::devices::DevicePresenceMonitor monitor{deviceGuid.value(), [](auto&& presenceEvent, auto&& deviceName) {
-        const auto presenceEventName = magic_enum::enum_name(presenceEvent);
-        std::cout << deviceName << " " << presenceEventName << std::endl;
-    }};
+    windows::devices::DevicePresenceMonitor monitor{ deviceGuid.value(),
+                                                     [](auto&& presenceEvent, auto&& deviceName) {
+                                                        const auto presenceEventName = magic_enum::enum_name(presenceEvent);
+                                                        std::cout << deviceName << " " << presenceEventName << std::endl;
+                                                     } 
+                                                    };
```
abeltrano, line 76: Is this change intentional?

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
forwardpointer: Oh I didn't know that plog::init returned a logger for which you could addAppender() to. IAppender by itself doesn't provide that function but the main logger initialized by plog::init does, so we could definitely save some lines of code here.

windows/tools/devicemon/Main.cxx
forwardpointer: The auto formatting was causing the tabbing to look weird so I changed it to this.

windows/shared/logging/logging-plog-windows/include/logging/windows/PlogDebugWrapper.hxx
forwardpointer: Ok I just pushed a change that removes the wrapper altogether because the plog::init function allows addAppender directly.

windows/tools/devicemon/Main.cxx
forwardpointer: I changed it back to how it was formatted before

windows/shared/logging/logging-plog-windows/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ target_include_directories(logging-plog-windows
 target_link_libraries(logging-plog-windows
     PUBLIC
         plog::plog
+        logutils
```
abeltrano, line 22: Why does this need to be `PUBLIC` instead of `PRIVATE`? Does some public part of this target/lib use something from `logutils`? The only public part of `logutils` seems to be `GetLogName()`, which I didn't notice being used here in the public interface of `logging-plog-windows`.

forwardpointer: Right now I have logging-plog-windows providing the tracelogging stuff as well as ```GetLogName```. I originally had it also provide the debug wrapper but I removed it, so now I guess I could have it not provide ```GetLogName`` and have all the windows code explicitly link with both the logutils library and the logging-plog-windows library. However, we might want to add something else to the windows specific logging library in the future so I think we should leave this structure.

abeltrano: > Right now I have logging-plog-windows providing the tracelogging stuff as well as GetLogName.

The tracelogging stuff, AFAIK, doesn't have anything in its public interface that uses anything from `logutils`. In that case, it should be linked as `PRIVATE`. If this changes later, then we can change to `PUBLIC` is appropriate.

> have all the windows code explicitly link with both the logutils library and the logging-plog-windows library

We should never have to explicitly link to any indirect dependencies; as long as we specify the link visibility (as here) correctly, CMake will take care of the rest. Otherwise, it becomes untenable figuring out all the dependencies in the chain of each lib.

abeltrano (changes requested):

windows/shared/logging/logging-plog-windows/CMakeLists.txt
forwardpointer: The tracelogging stuff also doesn't use the logutils library at all. I put it in PUBLIC so that linking against ```logging-plog-windows``` would provide both the logutils stuff and the tracelogging stuff.

abeltrano: I understand what the intention is, but `target_link_libraries` has a specific meaning: it transitively propagates the dependencies of the target. In this case, you're including another target as `PUBLIC` which declares a certain relationship, yet that relationship isn't reflective of the program structure since the target doesn't actually depend on `logutils`. Looking at that in isolation, it's confusing.

Despite that, if you're going to keep this, then I suggest adding a comment to clarify why it's being added as a `PUBLIC` transitive dependent when it's not actually a dependency so there is no confusion.

forwardpointer: I see, is there a way around this so it's not confusing, or is adding a comment the standard way?

abeltrano: I don't know of any way it's not confusing, but I've also not seen this done before (adding a non-dependent in the target link libraries). The standard approach is to only add direct, known dependencies to the target link libraries of a target.

forwardpointer: got it. In that case, I'll make the tracelogging stuff it's own library

abeltrano (approved):

PR #88, opened 2023-02-21
OnDeviceStatusChanged

forwardpointer:
### Type

- [ ] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* Implement the handler function OnDeviceStatusChanged
* Implement the conversion to neutral type for deviceStatus, genericError, sessionStatus
 
### Technical Details

* Added an Uninitialized value to the enum UwbDeviceState so we can print something different when the device is first initialized.
* Added some static const maps from ddi type to neutral type

### Test Results
None yet

### Checklist

### DIFF
--- a/lib/uwb/UwbDevice.cxx
+++ b/lib/uwb/UwbDevice.cxx
@@ -41,7 +41,15 @@ UwbDevice::OnStatusChanged([[maybe_unused]] UwbStatus status)
 void
 UwbDevice::OnDeviceStatusChanged([[maybe_unused]] UwbStatusDevice statusDevice)
 {
-    // TODO: implement this
+    if (m_status.State == UwbDeviceState::Uninitialized) {
+        m_status = statusDevice;
+        PLOG_VERBOSE << "initialized device status as " << m_status.ToString();
+        return;
+    }
+    if (m_status.State != statusDevice.State) {
+        PLOG_VERBOSE << "changed state: " << m_status.ToString() << " --> " << statusDevice.ToString();
+        m_status = statusDevice;
+    }
 }
 
 void
--- a/windows/devices/uwb/UwbCxAdapterDdiLrp.cxx
+++ b/windows/devices/uwb/UwbCxAdapterDdiLrp.cxx
@@ -534,10 +534,69 @@ windows::devices::uwb::ddi::lrp::To(const UWB_DEVICE_CAPABILITIES &deviceCapabil
     return uwbCapability;
 }
 
+static const std::unordered_map<UWB_DEVICE_STATE, UwbDeviceState> DeviceStateToMap{
+    { UWB_DEVICE_STATE_READY, UwbDeviceState::Ready },
+    { UWB_DEVICE_STATE_ACTIVE, UwbDeviceState::Active },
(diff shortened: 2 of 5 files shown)


lib/uwb/UwbDevice.cxx
```diff
@@ -41,7 +41,15 @@ UwbDevice::OnStatusChanged([[maybe_unused]] UwbStatus status)
...
 UwbDevice::OnDeviceStatusChanged([[maybe_unused]] UwbStatusDevice statusDevice)
 {
-    // TODO: implement this
+    if (m_status.State == UwbDeviceState::Uninitialized) {
+        m_status = statusDevice;
+        PLOG_VERBOSE << "initialized device status as " << m_status.ToString();
+        return;
+    }
+    if (m_status.State != statusDevice.State) {
+        PLOG_VERBOSE << "changed state: " << m_status.ToString() << " --> " << statusDevice.ToString();
```
abeltrano, line 50: nit: We have to `ToString()` helper such that you can do this instead without invoking `ToString()` directly:

```
#include <notstd/tostring.hxx>
using namespace strings::ostream_operators
```
...
```suggestion
    if (m_status.State == UwbDeviceState::Uninitialized) {
        m_status = statusDevice;
        PLOG_VERBOSE << "initialized device status as " << m_status;
        return;
    }
    if (m_status.State != statusDevice.State) {
        PLOG_VERBOSE << "changed state: " << m_status << " --> " << statusDevice;
```

forwardpointer: I tried your suggestion but it complained it couldn't find the operator<< for m_status :(

windows/devices/uwb/UwbCxAdapterDdiLrp.cxx
```diff
@@ -534,10 +534,95 @@ windows::devices::uwb::ddi::lrp::To(const UWB_DEVICE_CAPABILITIES &deviceCapabil
...
+    }
+    case UWB_NOTIFICATION_TYPE_GENERIC_ERROR: {
+        auto enumId = notstd::to_underlying(notificationData.genericError);
+        if (enumId < notstd::to_underlying(UWB_STATUS_ERROR_SESSION_NOT_EXIST)) {
+            return StatusToMapGeneric.at(notificationData.genericError);
+        }
+        if (enumId < notstd::to_underlying(UWB_STATUS_RANGING_TX_FAILED)) {
+            return StatusToMapSession.at(notificationData.genericError);
+        }
+        return StatusToMapRanging.at(notificationData.genericError);
```
abeltrano, line 613: We don't need a special case for this. We should just pass `notificationData.genericError` to the `::To(UWB_STATUS& status)` function which returns a `UwbStatus` (I suggested writing this function in another comment).

abeltrano (changes requested):

lib/uwb/UwbDevice.cxx
abeltrano: Hmm, probably the `PLOG` macro isn't actually an `ostream`. Oh well :'(

abeltrano (approved):
This is great.

PR #91, opened 2023-02-22
Session status

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* Implements OnSessionStatusChanged

### Developer Focus

* Added a m_sessionStatus private member that keeps track of the previous session status. Does it make sense to have the m_sessionStatus.SessionId possibly out of sync with the UwbSession::m_sessionId, especially upon initialization?

### DIFF
--- a/lib/uwb/UwbDevice.cxx
+++ b/lib/uwb/UwbDevice.cxx
@@ -72,7 +72,6 @@ UwbDevice::OnSessionStatusChanged(UwbSessionStatus statusSession)
         return;
     }
 
-    PLOG_VERBOSE << "session changed state:" << session->GetSessionStatus().ToString() << " --> " << statusSession.ToString();
     session->SetSessionStatus(statusSession);
 }
 
--- a/lib/uwb/UwbDevice.cxx
+++ b/lib/uwb/UwbDevice.cxx
@@ -61,6 +61,7 @@ UwbDevice::OnSessionStatusChanged(UwbSessionStatus statusSession)
         return;
     }
 
+    PLOG_VERBOSE << "session changed state:" << session->GetSessionStatus().ToString() << " --> " << statusSession.ToString();
     session->SetSessionStatus(statusSession);
 }
 
--- a/lib/uwb/UwbDevice.cxx
+++ b/lib/uwb/UwbDevice.cxx
@@ -61,7 +61,7 @@ UwbDevice::OnSessionStatusChanged(UwbSessionStatus statusSession)
         return;
     }
 
-    // TODO: implement this
+    session->SetSessionStatus(statusSession);
 }
 
 void
--- a/lib/uwb/UwbSession.cxx
+++ b/lib/uwb/UwbSession.cxx
@@ -74,6 +74,6 @@ UwbSession::StopRanging()
 void
 UwbSession::SetSessionStatus(const uwb::protocol::fira::UwbSessionStatus& status)
 {
-    PLOG_VERBOSE << "session changed state:" << m_sessionStatus.ToString() << " --> " << status.ToString();
+    PLOG_VERBOSE << "session changed state: " << m_sessionStatus.ToString() << " --> " << status.ToString();
     m_sessionStatus = status;
 }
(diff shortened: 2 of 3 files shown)


lib/uwb/UwbDevice.cxx
```diff
@@ -61,7 +61,8 @@ UwbDevice::OnSessionStatusChanged(UwbSessionStatus statusSession)
         return;
     }
 
-    // TODO: implement this
+    PLOG_VERBOSE << "session changed state:" << session->GetSessionStatus().ToString() << " --> " << statusSession.ToString();
```
abeltrano, line 75: This log would be better suited in `UwbSession::SetSessionStatus()` since then it will cover when the status is changed from _every_ source (eg. including unit test sources), instead of just from `UwbDevice`.

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -82,33 +82,48 @@ public:
...
+    void
+    SetSessionStatus(const uwb::protocol::fira::UwbSessionStatus&);
+
+    /**
+     * @brief Get the Session Status object
+     *
+     * @return const UwbSessionStatus&
+     */
+    const uwb::protocol::fira::UwbSessionStatus&
+    GetSessionStatus();
```
abeltrano, line 125: Is there any reason we need this right now? If there's no specific use, then let's remove it, especially since returning a reference to an object that can be asynchronously updated isn't thread safe. We can solve that problem, but I'd rather not introduce complexity now if we don't need it.

abeltrano (changes requested):

lib/uwb/include/uwb/UwbSession.hxx
forwardpointer: Generally speaking I like to make a getter for every setter but I agree that we don't need it right now and I didn't think that the function could be asynchronously called.

abeltrano: > Generally speaking I like to make a getter for every setter

I agree this is a good strategy in general.

> I didn't think that the function could be asynchronously called.

Here, the getter and setter can be called concurrently if the driver updates the session status while the service is trying to obtain the session status (say, to fulfill an API call).

lib/uwb/UwbSession.cxx
```diff
@@ -70,3 +70,10 @@ UwbSession::StopRanging()
         return;
     }
 }
+
+void
+UwbSession::SetSessionStatus(const uwb::protocol::fira::UwbSessionStatus& status)
+{
+    PLOG_VERBOSE << "session changed state:" << m_sessionStatus.ToString() << " --> " << status.ToString();
```
abeltrano, line 77: nit: there's no space between the `:` and the session value.
```suggestion
    PLOG_VERBOSE << "session changed state: " << m_sessionStatus.ToString() << " --> " << status.ToString();
```

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -82,33 +82,41 @@ public:
...
 
     /**
-     * @brief 
+     * @brief
      */
     void
     StopRanging();
 
+    /**
+     * @brief Set the Session Status object. NOTE, this function is NOT asynch safe
```
abeltrano, line 113: nit: thread-safe (coroutine for example, is concurrent but not async).
```suggestion
     * @brief Set the Session Status object. NOTE, this function is NOT thread-safe
```

lib/uwb/UwbSession.cxx
forwardpointer: The session id is already being logged by the UwbSessionStatus::ToString() function so I don't think we need to do it again.

abeltrano: 👍

abeltrano (approved):

PR #97, opened 2023-02-24
OnSessionRangingData

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Implement the handler for ranging data

### Technical Details

* Added another private variable UwbSessionData::m_rangingData that is just a vector of all observed ranging data.

### Reviewer Focus

Do we need a mutex for this handler?

### Checklist

### DIFF
--- a/lib/uwb/include/uwb/UwbSession.hxx
+++ b/lib/uwb/include/uwb/UwbSession.hxx
@@ -127,8 +127,8 @@ public:
 
     /**
      * @brief Add some ranging data to this session
-     * 
-     * @param data 
+     *
+     * @param data
      */
     void
     AddRangingData(const uwb::protocol::fira::UwbRangingData& data);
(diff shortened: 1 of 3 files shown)


lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -141,6 +149,7 @@ private:
 protected:
     uint32_t m_sessionId{ 0 };
     uwb::protocol::fira::UwbSessionStatus m_sessionStatus{};
+    std::vector<uwb::protocol::fira::UwbRangingData> m_rangingData{};
```
abeltrano, line 152: During a ranging session, we should expect a *lot* of these. Apart from invoking the session callback with this data, is there any reason we need to save this?

PR #103, opened 2023-02-28
Call the callback OnSessionMembershipChanged

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* Call the callback OnSessionMembershipChanged from within UwbSession::InsertPeers

### Technical Details

* UwbSession::InsertPeers is a templated function, taking in a view of UwbMulticastListStatus
* Swapped the dependency of UwbSession.hxx and UwbSessionEventCallbacks.hxx to make the template happy

### Reviewer Focus

* InsertPeers copies the relevant data to another vector, and then passes the vector to OnSessionMembershipChanged. 
* Maybe it makes sense to change the interface OnSessionMembershipChanged to also take in a range view so we don't need to copy?

### Checklist


lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -118,14 +92,39 @@ public:
...
+        std::vector<UwbPeer> uwbPeers;
+        {
+            std::scoped_lock peersLock{ m_peerGate };
+            for (const auto& peer : peers) {
+                InsertPeerImpl(peer.ControleeMacAddress);
+                uwbPeers.push_back(UwbPeer{ peer.ControleeMacAddress });
+            }
+        }
+        auto callbacks = m_callbacks.lock();
+        callbacks->OnSessionMembershipChanged(this, uwbPeers, std::vector<UwbPeer>{});
```
abeltrano, line 116: Need to check if the pointer is still valid.
```suggestion
        auto callbacks = m_callbacks.lock();
        if (callbacks != nullptr) {
            callbacks->OnSessionMembershipChanged(this, uwbPeers, std::vector<UwbPeer>{});
        }
```

abeltrano, line 116: nit: this can just be empty braces
```suggestion
        auto callbacks = m_callbacks.lock();
        callbacks->OnSessionMembershipChanged(this, uwbPeers, {});
```

lib/uwb/UwbSession.cxx
```diff
@@ -32,11 +32,7 @@ UwbSession::AddPeer(UwbMacAddress peerMacAddress)
 {
     std::scoped_lock peersLock{ m_peerGate };
     PLOG_VERBOSE << "adding peer with address " << peerMacAddress.ToString();
-    auto [_, inserted] = m_peers.insert(peerMacAddress);
-
-    if (inserted) {
-        AddPeerImpl(std::move(peerMacAddress));
-    }
+    AddPeerImpl(std::move(peerMacAddress));
     PLOG_VERBOSE << "peer added";
```
abeltrano, line 36: Since you're here, can you modify the peer-related plog prints to ensure the session id is included in the message?

lib/uwb/UwbSession.cxx
```diff
@@ -32,11 +32,7 @@ UwbSession::AddPeer(UwbMacAddress peerMacAddress)
 {
     std::scoped_lock peersLock{ m_peerGate };
     PLOG_VERBOSE << "adding peer with address " << peerMacAddress.ToString();
```
abeltrano, line 34: Since you're here, can you modify the peer-related plog prints in this file to include the session id in a consistent message?
```suggestion
    PLOG_VERBOSE << "Session with id " << m_sessionId << " added peer with address " << peerMacAddress.ToString();
```

lib/uwb/UwbSession.cxx
abeltrano, line 36: This can be removed since there is a plog on line 34, right?

abeltrano (changes requested):

abeltrano:
This was really well done.

PR #104, opened 2023-02-28
Handle ranging data notification

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* Calls the OnPeerPropertiesChanged callback when receiving new ranging data

### Technical Details

* Adds a conversion function from Q9.7 format to standard IEEE double format.

### Test Results

None yet

### Reviewer Focus

Test will be added for the conversion function

### Checklist


lib/uwb/UwbPeer.cxx
```diff
@@ -32,6 +33,47 @@ UwbPeer::UwbPeer(UwbMacAddress address) :
     m_address(std::move(address))
 {}
 
+/**
+ * @brief Assuming the Arm definition of Qm.n formatting, the most significant bit is the sign, the next
+ * (m-1) bits are an integer, and the next n bits is the number to be multiplied by pow(2,n)
+ * The double equivalent will be the sum of those two results
+ * TODO double check this conversion, write tests for it
```
abeltrano, line 40: This description should go below the `@brief`.

```suggestion
 * @brief Converts a Q9.7-formatted value to an IEEE 754 double precision floating pointer formatted value.
 *
 *  Assuming the Arm definition of Qm.n formatting, the most significant bit is the sign, the next
 * (m-1) bits are an integer, and the next n bits is the number to be multiplied by pow(2,n)
 * The double equivalent will be the sum of those two results
 * TODO double check this conversion, write tests for it
```

lib/uwb/UwbPeer.cxx
```diff
@@ -32,6 +33,47 @@ UwbPeer::UwbPeer(UwbMacAddress address) :
...
+ *
+ * @param q97 a number in Q9.7 format
+ * @return double
+ */
+double
+ConvertQ9_7FormatToIEEE(uint16_t q97)
```
abeltrano, line 46: Please convert to camelCase.

abeltrano, line 46: Probably this should go into some kind of helper class or library, but we can leave it here for now (so feel free to resolve this).

lib/uwb/UwbSession.cxx
```diff
@@ -78,4 +78,14 @@ UwbSession::InsertPeerImpl(const uwb::UwbMacAddress& peerAddress)
 {
     m_peers.insert(peerAddress);
     PLOG_VERBOSE << "Session with id " << m_sessionId << " added peer via DDI with mac address " << peerAddress.ToString();
+}
+
+void
+UwbSession::ProcessRangingData(const std::vector<uwb::UwbPeer>& peerRangingData)
+{
+    auto callbacks = m_callbacks.lock();
+    if (!callbacks) {
```
abeltrano, line 87: I think the logic here is flipped.
```suggestion
    if (callbacks != nullptr) {
```

lib/uwb/include/uwb/UwbPeer.hxx
```diff
@@ -39,6 +44,8 @@ public:
      */
     explicit UwbPeer(UwbMacAddress address);
 
+    explicit UwbPeer(const uwb::protocol::fira::UwbRangingMeasurement& data);
```
abeltrano, line 47: Please add a doxygen comment header here.

lib/uwb/UwbDevice.cxx
```diff
@@ -111,7 +111,15 @@ UwbDevice::OnSessionRangingData(UwbRangingData rangingData)
...
     }
 
-    // TODO: implement this
+    PLOG_VERBOSE << "Session with id " << rangingData.SessionId << " processing new ranging data";
+    std::vector<UwbPeer> peersData;
+    peersData.reserve(rangingData.RangingMeasurements.size());
+    for (const auto& peerData : rangingData.RangingMeasurements) {
+        UwbPeer data{ peerData };
+        PLOG_VERBOSE << "Peer data: " << data.ToString();
+        peersData.emplace_back(data);
```
abeltrano, line 120: This should really use `push_back` instead, sine a copy is actually being made here. `emplace_back()` is intended for in-place pushing. That could be done here using `emplace_back(peerData)` and avoid creating the `UwbPeer` data object in the loop, but since you're logging it, what you've done makes sense.
```suggestion
        peersData.push_back(std::move(data));
```

lib/uwb/UwbPeer.cxx
```diff
@@ -32,6 +33,47 @@ UwbPeer::UwbPeer(UwbMacAddress address) :
...
+    m_spatialProperties{
+        .Distance{ data.Distance }, // TODO is this also q97
+        .AngleAzimuth{ ConvertQ9_7FormatToIEEE(data.AoAAzimuth.Result) },
+        .AngleElevation{ ConvertQ9_7FormatToIEEE(data.AoAElevation.Result) },
+        .Elevation{ ConvertQ9_7FormatToIEEE(data.AoaDestinationElevation.Result) }, // TODO is this right?
+
+        .AngleAzimuthFom{ data.AoAAzimuth.FigureOfMerit },
+        .AngleElevationFom{ data.AoAElevation.FigureOfMerit },
+        .ElevationFom{ data.AoaDestinationElevation.FigureOfMerit }
+    }
```
abeltrano, line 73: Might be worth adding a c'tor to `UwbPeerSpatialProperties` that takes a `uwb::protocol::fira::UwbRangingMeasurement` as an input argument.

abeltrano (changes requested):

lib/uwb/UwbPeer.cxx
forwardpointer: Should it be PascalCase instead? Also, I'm unsure how to format Q9.7 in PascalCase. I guess I could just do Q97 and hope users know that it doesn't mean a number with 2**(-97) precision.

forwardpointer: Agreed.

lib/uwb/UwbSession.cxx
forwardpointer: Nice catch!

lib/uwb/UwbPeer.cxx
abeltrano: Yes sorry, PascalCase for function names. All that really matters is it starts with a capital and doesn't use "snakes", so `ConvertQ97FormatToIEEE` is good.

lib/uwb/UwbPeer.cxx
forwardpointer: I think it makes sense to hold off on implementing the reverse conversion because we may want to put all of the conversion functions in the same place to avoid circular header dependencies

abeltrano (dismissed):

forwardpointer:
#107

abeltrano (approved):

PR #123, opened 2023-03-06
User/corbinphipps/add individual command capability

corbin-phipps:
### Type

- [x] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR adds functionality to the nocli tool to send raw individual commands to the driver, specifically the GetDeviceInfo command. The tool also outputs the data returned by the device/driver stack.

### Technical Details

Added "raw" subcommand to nocli.
Added "getdeviceinfo" subcommand under the "raw" subcommand in nocli.
Added "Initialize" virtual function to UwbDevice base class so nocli can initialize the device.
Fixes the UwbVersion ToString() method to print the numbers instead of strange symbols.

### Test Results

The nocli tool successfully gets the device information from the TestClient driver and simulated device.

### Reviewer Focus

Structure of added subcommands; Initialize() virtual function in UwbDevice base class

### Future Work

Print out VendorSpecificInfo as blob. Could potentially print in readable format if we know what that format is (which we do not know)

### Checklist


lib/uwb/UwbDevice.cxx
```diff
@@ -198,6 +198,14 @@ UwbDevice::Reset()
     ResetImpl();
 }
 
+bool
+UwbDevice::Initialize()
+{
+    // TODO: Either make this do something useful, or make function pure virtual
+    // and have every derived class implement it
+    return true;
```
abeltrano, line 206: It seems reasonable that not all derived classes will require initialization, so this makes sense as it is since this will be the default implementation if not override. You can remove the TODO.

lib/uwb/UwbVersion.cxx
```diff
@@ -14,7 +14,7 @@ std::string
 UwbVersion::ToString() const
 {
     std::ostringstream ss;
-    ss << Major << '.' << Minor << '.' << Maintenance;
+    ss << +Major << '.' << +Minor << '.' << +Maintenance;
```
abeltrano, line 17: Thanks for fixing my bug!

lib/uwb/protocols/fira/FiraDevice.cxx
```diff
@@ -106,10 +106,17 @@ UwbDeviceInformation::ToString() const
 {
     std::ostringstream ss;
     if (IsUwbStatusOk(Status)) {
+        ss << "Status: " << ::ToString(Status) << "\n";
```
abeltrano, line 109: This will just print 'Ok' every time, I don't think we need it. In general, we've been excluding the status from structures once they've been decoded since the status isn't (in most cases) part of the state, but instead, describes whether obtaining the state was successful or not.

lib/uwb/include/uwb/UwbDevice.hxx
```diff
@@ -51,6 +51,12 @@ public:
     void
     Reset();
 
+    /**
+     * @brief Initializes a new UWB device
+     */
+    virtual bool
+    Initialize();
```
abeltrano, line 58: Looks like this doxygen comment needs updating to account for `bool` return type.

tools/cli/NearObjectCli.cxx
```diff
@@ -292,3 +317,30 @@ NearObjectCli::AddSubcommandUwbRangeStop(CLI::App* parent)
...
+        if (!uwbDevice) {
+            std::cerr << "no device found" << std::endl;
+            return;
+        }
+        
+        if (uwbDevice->Initialize()) {
+            m_cliHandler->HandleGetDeviceInfo(uwbDevice);
+        } else {
+            std::cerr << "device not initialized" << std::endl;
+        }
```
abeltrano, line 342: To keep with the prevailing pattern, I think we should return early when an error is recognized to redeuce unnecessary nested scoping.

```suggestion
        if (!uwbDevice->Initialize()) {
            std::cerr << "device not initialized" << std::endl;
            return;
        }
        
        m_cliHandler->HandleGetDeviceInfo(uwbDevice);
```

lib/uwb/protocols/fira/FiraDevice.cxx
corbin-phipps: I see, that seems perfectly reasonable to me. I'll remove this line

lib/uwb/include/uwb/UwbDevice.hxx
corbin-phipps: Yes, good catch!

lib/uwb/include/uwb/UwbDevice.hxx
```diff
@@ -51,6 +51,14 @@ public:
     void
     Reset();
 
+    /**
+     * @brief Initializes a new UWB device
+     * 
+     * @return bool
+     */
+    virtual bool
+    Initialize();
```
abeltrano, line 60: This should follow the existing pimpl pattern where the top-level `UwbDevice` object defines the interface behavior with a non-virtual function (here, `bool Initialize()`) and then declares a private virtual impl function that that derived classes implement (here, `bool InitializeImpl`). If the derived class is required to provide such an impl, then it should be pure virtual (`eg. bool InitializeImpl() = 0`), but in this case it makes sense not to make it pure since some derived impls wouldn't need to initialize anything.

This pattern helps keep the interface stable and allows us to instrument operations from the highest level in the class hierarchy.

abeltrano (changes requested):
Just a few small tweaks, then I think this is good to go.

PR #124, opened 2023-03-07
User/corbinphipps/nocli add reset

corbin-phipps:
### Type

- [x] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR adds the device reset command to the nocli tool

### Technical Details

Added "devicereset" sub-command to nocli
Fix the device reset code in UwbDeviceConnector to properly call DeviceIoControl

### Test Results

Device reset succeeds on TestClient driver

### Reviewer Focus

None.

### Future Work

None.

### Checklist

### DIFF
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -80,6 +80,12 @@ NearObjectCli::GetRangeStopApp() noexcept
 }
 
 CLI::App&
+NearObjectCli::GetRawDeviceResetApp() noexcept
+{
+    return *m_rawDeviceResetApp;
+}
+
+CLI::App&
 NearObjectCli::GetRawGetDeviceInfoApp() noexcept
 {
     return *m_rawGetDeviceInfoApp;
@@ -216,6 +222,7 @@ NearObjectCli::AddSubcommandUwbRaw(CLI::App* parent)
     auto rawApp = parent->add_subcommand("raw", "individual commands")->require_subcommand()->fallthrough();
 
     // sub-commands
+    m_rawDeviceResetApp = AddSubcommandUwbRawDeviceReset(rawApp);
     m_rawGetDeviceInfoApp = AddSubcommandUwbRawGetDeviceInfo(rawApp);
 
     return rawApp;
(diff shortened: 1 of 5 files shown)


tools/cli/NearObjectCli.cxx
```diff
@@ -79,6 +79,12 @@ NearObjectCli::GetRangeStopApp() noexcept
     return *m_rangeStopApp;
 }
 
+CLI::App&
+NearObjectCli::GetRawDeviceResetApp() noexcept
+{
+    return *m_rawDeviceResetApp;
+}
+
```
abeltrano, line 87: For individual commands that are terminal (ie. don't have sub-commands), then I don't think there's any need to save away the `CLI::App` instance pointer. The `AddSubcommandUwbRawDeviceReset` function can be modified to throw if there's an error, or, have the caller check the pointer for non-`nullptr` and handle the failure appropriately (exiting the program would be fine). 

So, you can remove this function and the `m_rawDeviceResetApp` member. Same goes for `m_rawGetDeviceInfoApp` for the `devicegetinfo` command, I should have called this out on the prior PR.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -47,17 +47,34 @@ UwbDeviceConnector::Reset()
...
+    auto reset = reinterpret_cast<UWB_DEVICE_RESET*>(std::data(deviceResetInputBuffer));
+
+    reset->size = deviceResetInputBufferLength;
+    reset->resetConfig = UWB_RESET_CONFIG_UWBS_RESET;
+
+    std::vector<uint8_t> deviceResetStatusBuffer{};
+    DWORD bytesRequired = sizeof(UWB_STATUS);
+    deviceResetStatusBuffer.resize(bytesRequired);
+
+    BOOL ioResult = DeviceIoControl(handleDriver.get(), IOCTL_UWB_DEVICE_RESET, std::data(deviceResetInputBuffer), deviceResetInputBufferLength, std::data(deviceResetStatusBuffer), std::size(deviceResetStatusBuffer), &bytesRequired, nullptr);
```
abeltrano, line 61: Neither the `vector` for either arg nor the `bytesRequired` arg should be required here since the sizes of both args are fixed. Can this work?

```suggestion
    const UWB_DEVICE_RESET deviceReset {
        .size = sizeof(UWB_DEVICE_RESET),
        .resetConfig = UWB_RESET_CONFIG_UWBS_RESET,
    };
    UWB_STATUS status;
    BOOL ioResult = DeviceIoControl(handleDriver.get(), IOCTL_UWB_DEVICE_RESET, const_cast<UWB_DEVICE_RESET *>(&deviceReset), sizeof deviceReset, &status, sizeof status, nullptr, nullptr);

```

abeltrano (changes requested):

tools/cli/NearObjectCli.cxx
corbin-phipps: Good point, I'll remove these

windows/devices/uwb/UwbDeviceConnector.cxx
corbin-phipps: Ah I thought bytesRequired was necessary for DeviceIoControl, but that's only if the "overlapped" field is not null, which in this case it is. Your suggested change works so I will push the changes

abeltrano (approved):

forwardpointer (approved):

Issue #96, opened 2023-02-24
Make the InsertPeer function more restrictive in what can call it


forwardpointer:
Right now, the UwbSession::InsertPeer function is public, meaning anyone can call it. Ideally, we'd want to make it so that only the UwbDevice can call it.

PR #125, opened 2023-03-08
Change the IO relationship between UwbDevice and UwbSession to go through UwbDeviceConnector

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [x] Feature update
- [x] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* UwbDevice publishes its UwbDeviceEventCallbacks to UwbDeviceConnector, and likewise UwbSession publishes its UwbSessionEventCallbacks to UwbDeviceConnector
* UwbDeviceConnector is now responsible for deciding which callback to call given UwbNotificationData
** This is implemented in UwbDeviceConnector.cxx:UwbDeviceConnector::HandleNotifications
* Changed the interfaces UwbSessionEventCallbacks and UwbDeviceEventCallbacks to simply be structs of lambdas, instead of pure virtual interfaces. This could allow users of the interface to more easily instantiate callbacks to pass around.
* Changed the interface UwbSessionEventCallbacks to not require the UwbSession *pointer. Instead it will be captured in the lambda closure.
** This means that instead of passing the callbacks to the UwbSession when you first create it, instead you pass it in when you configure the UwbSession. This is reflected in tools/cli/NearObjectCliHandler.cxx

### Test Results

I've verified it compiles fine on Windows. I will do the nocli uwb start range command as a sanity check.

### Reviewer Focus

The reason that I made the UwbSessionEventCallbacks interface capture the UwbSession *pointer in the lambda closure was because UwbSessionEventCallbacks needs to somehow lookup the UwbSession in order to execute the callback. There three alternatives that I considered.
1. If I leave the interface as it is, I could either make UwbDeviceConnector a friend of UwbDevice, or the other way around, since the GetSession function is a private member of UwbDevice. However this would either mean 
   * that UwbDevice would have access to the private member functions of UwbDeviceConnector that handle talking to the DDI, which would render the Connector pointless,
   * or that UwbDeviceConnector would have access to the ability to functions like InitializeImpl, which would also be undesireable
2. I could also have UwbDeviceConnector maintain a set of UwbSession so it will have its own private member function GetSession(). However, this means that now either we have two sets of UwbSession (which would be terrible to maintain consistency), or we no longer have UwbDevice maintaining a set of UwbSession, and instead only have UwbDeviceConnector doing so. If we do the latter, this means that we would somehow have to notify UwbDeviceConnector every time UwbDevice tries to CreateSession, which essentially means that one of the two classes needs to be friends of the other.
3. The last option I considered was to add another struct PreparedUwbSessionEventCallbacks that esentially is just UwbSessionEventCallbacks, but prepared with the lambda closure. The idea is that upon construction, UwbSession takes in UwbSessionEventCallbacks and turns it into PreparedUwbSessionEventCallbacks, which it then stores. This solves the problem of the interface change affecting a lot of files, but also introduces the question of who owns the event callbacks. Originally, whoever created the UwbSession would be in charge of cleaning up the callback, but now UwbSession is in charge of cleaning it up. This may not be that bad, and so I am leaving this option open for the future.

### Future Considerations
Maybe it makes sense to make the UwbDeviceConnector have a redefined role as "the stuff that actually talks to the driver, as well as the stuff that dispatches the callbacks". This would mean that aside from the DDI specific stuff, much of the code for UwbDeviceConnector could be pushed to the generic implementation and available for the linux side. I believe this makes sense because right now there is nothing that invokes the UwbDevice event callbacks in the base implementation and hence the linux side.

### Checklist


forwardpointer:
https://github.com/microsoft/nearobject-framework/issues/96

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -43,16 +42,17 @@ public:
     GetId() const noexcept;
 
     /**
-     * @brief Configure the session for use.
+     * @brief Configure the session for use. Also moves the callbacks into ownership here
```
forwardpointer, line 45: I'm not entirely sure about the behavior of shared/weak pointers, but my understanding is that if a shared pointer is passed into this function, the ownership actually does NOT transfer.

forwardpointer:
related:
https://github.com/microsoft/nearobject-framework/issues/126

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -335,3 +444,15 @@ UwbDeviceConnector::NotificationListenerStop()
...
 }
+
+void
+UwbDeviceConnector::RegisterDeviceEventCallbacks(::uwb::UwbDeviceEventCallbacks callbacks)
+{
+    m_deviceEventCallbacks = callbacks;
+}
+
+void
+UwbDeviceConnector::RegisterSessionEventCallbacks(uint32_t sessionId, std::weak_ptr<::uwb::UwbSessionEventCallbacks> callbacks)
```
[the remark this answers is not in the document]
forwardpointer: The reason for weak_ptr on session callback is that they capture the ```this``` pointer in the lambda closure, meaning that their lifetime is completely tied to the lifetime of the UwbSession object. Thus I had the UwbSession object own the lambda, which means that UwbDeviceConnector is merely borrowing them via weak_ptr.

Regarding the device callbacks, I'm not entirely certain what the best way to approach it is, but probably using weak pointers for that would also make sense.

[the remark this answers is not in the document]
abeltrano: To give a little color here: `UwbSession` was originally designed to allow the caller to completely control the lifetime of the callbacks. Whether or not the `UwbSession` has a set of registered callbacks has no influence on its operation, so it doesn't need to keep those callbacks alive with a strong reference. `weak_ptr` solves this perfectly since it does not itself repreent a strong reference until it is promoted (via a successful `lock()` call) to a `shared_ptr`. This way, `UwbSession` takes and holds a strong reference only for as long as it takes to invoke the callback(s) in question.

Another thing to note is that, capturing `this` in a lambda capture doesn't affect the object's (described by `this`) lifetime or ownership in any way. It simply copies the `this` pointer into the capture. Think of a lambda being implemented like this:

```
Object theObject = new Obejct(...);
void Object::LambdaExample()
{
    int argByRef = 10;
    int argByValue = 20;    
    auto func = [this, &argByRef, argByValue](){};
    m_func = std::move(func); // store the lambda for later use
}

// the compiler generates something like this for `func` above: 
struct FuncLambda {
    Object *pThis = theObject;
    int& argByRef = argByRef;
    int argByValue = argByValue;
    // fancy stuff storing the actual lambda function code and args (none here for simplicity)
};
```
This is really bad code because the lambda captures a function-local variable, `argByRef`, by reference. Once `LambdaExample()` finishes running, the memory for `funcByRef` no longer exists, yet the lambda references it. This is obviously really bad since the lambda will silently access invalid memory if invoked after the function completes. The same applies to the `this` pointer here: if that lambda was passed to another object and invoked after `theObject` is deleted, the same problem arises.

So, if you're capturing the `this` pointer in a lambda that another object is storing, you absolutely need to ensure that the captured object is not destroyed before that lambda is used. The two object's lifetimes are thus co-dependent. Unless the lifetime of the thing capturing the `this` pointer is directly tied to the lifetime of the thing storing the lambda (ie. one object wholly owns the other), then the safest way to enforce this is with strong reference counting (ie. `shared_ptr`).

All that said, I think using a `weak_ptr` here for the connector session callbacks could be a valid solution, though it does highlight the additional challenges involved using lambdas vs class instances. With a class instance represented though a (`weak/shared/unique`) pointer, it's impossible for the callbacks to reference invalid memory since the callbacks themselves are members of the class which we know is valid (via the ref count of the shared/weak/unique) pointer. With the lambda, it could have captured anything from anywhere, and even a `weak_ptr` to the class containing the lambdas doesn't tell us anything about the validity of the callbacks themselves. That doesn't mean it won't work, only that we get a less strong guarantee about their behavior. In a vacuum, I tend to prefer the class instance method since it makes the code more robust and less prone to programmer error.

Using a value-copy for the device callbacks is less clear to me, since this implies that the source of the device callbacks will always exist as long as the connector exists, and I don't think this is true, or even if it is true, it should not have to be. Any code should be able to create a connector and reigster device callbacks, not just code in `UwbDevice`.

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -48,14 +50,31 @@ public:
...
-    NotificationListenerStart(std::function<void(::uwb::protocol::fira::UwbNotificationData)> onNotification);
+    NotificationListenerStart();
 
     /**
      * @brief Stop listening for notifications.
      */
     void
     NotificationListenerStop();
 
+    /**
```
abeltrano, line 61: I really like this interface, it's simple and accomplishes exactly what we need.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -308,13 +309,121 @@ UwbDeviceConnector::HandleNotifications(wil::shared_hfile handleDriver, std::sto
...
+    auto it = m_sessionEventCallbacks.find(sessionId);
+    if (it == std::end(m_sessionEventCallbacks)) {
+        PLOG_WARNING << "Ignoring MulticastListStatus event due to missing session callback";
+        return;
+    }
+
+    auto callbacks = it->second.lock();
+    if (not(callbacks->OnSessionMembershipChanged)) {
+        PLOG_WARNING << "Ignoring MulticastListStatus event due to missing session callback";
+        return;
```
abeltrano, line 340: In this case, we should remove the callback from `m_sessionEventCallbacks` since it can never be valid again. `m_sessionEventCallbacks.erase(it);` should do the trick.

abeltrano:
> ### Future Considerations
> Maybe it makes sense to make the UwbDeviceConnector have a redefined role as "the stuff that actually talks to the driver, as well as the stuff that dispatches the callbacks". This would mean that aside from the DDI specific stuff, much of the code for UwbDeviceConnector could be pushed to the generic implementation and available for the linux side. I believe this makes sense because right now there is nothing that invokes the UwbDevice event callbacks in the base implementation and hence the linux side.

This is a good observation; we'll have to think about this a bit more. What you have here is good for now, though.

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -48,14 +50,31 @@ public:
...
+    void
+    RegisterSessionEventCallbacks(uint32_t sessionId, std::weak_ptr<::uwb::UwbSessionEventCallbacks> callbacks);
+
+    /**
+     * @brief Sets the callbacks for the UwbDevice that owns this UwbDeviceConnector
+     *
+     * @param callbacks
+     */
+    void
+    RegisterDeviceEventCallbacks(::uwb::UwbDeviceEventCallbacks callbacks);
```
abeltrano, line 76: I really like this interface, it's simple and accomplishes exactly what we need.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -308,13 +309,121 @@ UwbDeviceConnector::HandleNotifications(wil::shared_hfile handleDriver, std::sto
...
+UwbDeviceConnector::OnSessionMulticastListStatus(::uwb::protocol::fira::UwbSessionUpdateMulicastListStatus statusMulticastList)
+{
+    uint32_t sessionId = statusMulticastList.SessionId;
+    auto it = m_sessionEventCallbacks.find(sessionId);
+    if (it == std::end(m_sessionEventCallbacks)) {
+        PLOG_WARNING << "Ignoring MulticastListStatus event due to missing session callback";
+        return;
+    }
+
+    auto callbacks = it->second.lock();
```
abeltrano, line 337: nit: I find using a structured binding on `std::pair` and `std::tuple` makes the code a bit easier to understand. For example, I have to do some work to figure out what `second` is here, whereas a structured binding makes it clear:
```suggestion
        auto& [_, callbacksWeak] = it; 
        auto callbacks = it->callbacksWeak.lock();
```

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
abeltrano, line 76: One thing missing here is the ability to unregister the callbacks. When a `weak_ptr` is passed, the un-registration can be implicit since the caller can simply delete the object containing the callbacks and the connector then won't be able to resolve them and can remove its copy of the `weak_ptr`. Even in that case, though, it would be better to have a solution that allowed explicit deregistration to support the scenario where the callbacks are provided in a class that implements multiple interfaces simultaneously, eg.:

```
struct UberUwbSessionImpl : 
    public UwbDeviceEventCallbacks, public UwbSessionEventCallbacks, public ISomeOtherInterface
{
    // impl here
};
```

One way to allow un-registration is to return some kind of opaque token that the implementation uses to associate the callbacks with the registration, then the caller uses the token to un-register its callbacks. This can be done with some kind of unique identifier, an opaque pointer, or some combination thereof.

lib/uwb/include/uwb/UwbSessionEventCallbacks.hxx
```diff
@@ -41,67 +41,51 @@ enum class UwbSessionEndReason {
...
  * method to receive information from near peers.
- * 
- * A pointer to the UwbSession for which the callbacks are bound is provided in
- * each callback. This enables using the same callback object instance with
- * multiple sessions where the pointer value can be used to lookup the
- * associated session for the event. This pointer must not be cached nor used
- * beyond execution of the callback.
+ *
  */
 struct UwbSessionEventCallbacks
```
abeltrano, line 46: Omitting the `UwbSession` pointer sacrifices the caller's ability to specify a single set of callbacks for all instances they wish to register since the session pointer disambiguates which session the callback is for. We should restore this argument in all the callback functions here.

This goes back to the design challenge you faced earlier, in that the connector has no knowledge of a `UwbSession` (which is good) so it can't easily determine the `UwbSession*` pointer to pass. The dilemma of ownership of lambdas is yet another design challenge and complication. Together, I think these issues reveal that the callbacks invoked by the `UwbDeviceConnector` should be a completely separate entity from those invoked by `UwbSession` and `UwbDevice` since they're providing different functionality. The connector deals in session ids, the `UwbDevice` and `UwbSession` deal in `UwbSession` object instances.

I think the way you've defined the callbacks now (with lambdas) are perfect for `UwbDeviceConnector` because `UwbDeviceConnector` is always meant to be wholly owned by something, and the way they were defined previously were suitable for `UwbDevice` and `UwbSession`. So, I think the best solution is to restore the old callbacks, and take the new ones you've defined and make them specific to `UwbDeviceConnector`.

abeltrano (changes requested):

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -106,9 +125,35 @@ private:
...
+    void
+    OnSessionMulticastListStatus(::uwb::protocol::fira::UwbSessionUpdateMulicastListStatus statusMulticastList);
+
+    /**
+     * @brief Internal function that prepares the notification for processing by the m_sessionEventCallbacks
+     *
+     * @param rangingData
+     */
+    void
+    OnSessionRangingData(::uwb::protocol::fira::UwbRangingData rangingData);
```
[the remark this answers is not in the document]
forwardpointer: That's right, these are invoked internally only, and they are ```private```

windows/devices/uwb/UwbDeviceConnector.cxx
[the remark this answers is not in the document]
forwardpointer: I think I'm going to use a weak_ptr for the device callbacks

lib/uwb/include/uwb/UwbSessionEventCallbacks.hxx
forwardpointer: I reverted the UwbDeviceEventCallbacks and UwbSessionEventCallbacks to how they were before and added the UwbRegisteredCallbacks which will what the UwbDeviceConnector use to execute those callbacks.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -17,6 +18,14 @@ using namespace windows::devices;
 using namespace windows::devices::uwb;
 using namespace ::uwb::protocol::fira;
 
+namespace windows::devices::uwb
+{
+class RegisteredCallbackToken
+{
+    uint32_t callbackId;
+};
+}; // namespace windows::devices::uwb
```
abeltrano, line 27: This semi-colon is extraneous:

```suggestion
} // namespace windows::devices::uwb

```

windows/devices/uwb/UwbSession.cxx
```diff
@@ -17,7 +17,50 @@ using namespace ::uwb::protocol::fira;
...
+        },
+        [this](const std::vector<::uwb::UwbPeer> peersAdded, const std::vector<::uwb::UwbPeer> peersRemoved) {
+            auto callbacks = m_callbacks.lock();
+            if (not callbacks) {
+                PLOG_WARNING << "missing session event callback for ranging data, skipping";
+                // TODO deregister
+            }
+            return callbacks->OnPeerPropertiesChanged(this, peersAdded);
+        });
+    m_uwbDeviceConnector->RegisterSessionEventCallbacks(m_sessionId, m_registeredCallbacks);
```
abeltrano, line 62: This needs to save the returned `RegisteredCallbackToken *` for later deregistration, correct?

abeltrano (changes requested):
This is getting pretty close, the updates here are great.

windows/devices/uwb/UwbDevice.cxx
```diff
@@ -101,10 +112,8 @@ bool
 UwbDevice::InitializeImpl()
 {
     m_uwbDeviceConnector = std::make_shared<UwbDeviceConnector>(m_deviceName);
-    m_uwbDeviceConnector->NotificationListenerStart([this](auto&& uwbNotificationData) {
-        // Invoke base class notification handler which takes care of threading.
-        ::UwbDevice::OnUwbNotification(std::move(uwbNotificationData));
-    });
+    m_uwbDeviceConnector->RegisterDeviceEventCallbacks(m_callbacks);
```
abeltrano, line 115: This needs to save the registration token, right?

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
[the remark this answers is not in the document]
abeltrano: Apologies, the review context had `public` just before the snippet and didn't show the real `private:` access modifier, so I mistook them for `public`.

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
forwardpointer: I added a skeleton for deregistration. I think it makes the most sense for session callbacks, as sessions come and go but there is only one device that owns a UwbDeviceConnector so we might not need it for device callbacks. We can figure more details out in the future.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -308,13 +317,152 @@ UwbDeviceConnector::HandleNotifications(wil::shared_hfile handleDriver, std::sto
...
+    auto it = m_sessionEventCallbacks.find(sessionId);
+    if (it == std::end(m_sessionEventCallbacks)) {
+        PLOG_WARNING << "Ignoring RangingData event due to missing session callback";
+        return;
+    }
+    auto callbacks = it->second.lock();
+    if (not(callbacks->OnPeerPropertiesChanged)) {
+        PLOG_WARNING << "Ignoring RangingData event due to missing session callback";
+        return;
+    }
```
abeltrano, line 387: Can you replicate the approach used in `OnSessionMulticastListStatus` here? That used a structured binding for the map callbacks and erased the callbacks from the map when `lock()` returned a `nullptr`.

forwardpointer: Thanks, I forgot about this one.

abeltrano (approved):

PR #141, opened 2023-03-15
Fix --help output for range start

corbin-phipps:
### Type

- [x] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR fixes a bug with the nocli tool where the output of "nocli.exe uwb range start --help" would print out random symbols for enum values of type uint8_t.

### Technical Details

Because the code that prints the enum values is in the CLI11 library, we have no control over it and cannot fix the problem if we supply it uint8_t enum values. Therefore, this PR eliminates the call to CLI::CheckedTransformer() and creates a string with the enum usage to be printed as the option description.

### Test Results

The output is clean and has no random symbols.

### Reviewer Focus

None.

### Future Work

By nature of std::unordered_map, there is no guarantee in the order in which the enum values are printed when using --help. This doesn't actually matter but it would be nice to have values printed in order (e.g. for --Channel: C5, C6, C8, ...). However, simply switching to std::map doesn't work because of situations where numbers are represented as strings (e.g. 10 would come before 5).

### Checklist

### DIFF
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -109,20 +109,20 @@ namespace detail
  * corresponding values.
  *
  * @tparam EnumType The type of the enumeration.
- * @return std::map<std::string, int>
+ * @return std::unordered_map<std::string, EnumType>
  */
 template <typename EnumType>
 // clang-format off
 requires std::is_enum_v<EnumType>
-const std::unordered_map<std::string, int>
+const std::unordered_map<std::string, EnumType>
 CreateEnumerationStringMap() noexcept
 // clang-format on
 {
     const auto& reverseMap = magic_enum::enum_entries<EnumType>();
-    std::unordered_map<std::string, int> map;
+    std::unordered_map<std::string, EnumType> map;
 
     for (const auto& [enumValue, enumName] : reverseMap) {
-        map.emplace(enumName, static_cast<int>(enumValue));
+        map.emplace(enumName, enumValue);
     }
 
     return map;
@@ -148,13 +148,15 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo)
 {
     std::string optionName{ std::string("--").append(magic_enum::enum_type_name<EnumType>()) };
 
-    const auto& map = CreateEnumerationStringMap<EnumType>();
+    const auto map = CreateEnumerationStringMap<EnumType>();
     std::ostringstream enumUsage;
 
     auto it = std::cbegin(map);
-    enumUsage << "value in { " << it->second << "(" << it->first << ")";
-    for (it = std::next(std::cbegin(map)); it != std::cend(map); ++it) {
-        enumUsage << ", " << it->second << "(" << it->first << ")";
+    const auto& [name, value] = *it;
+    enumUsage << "value in { " << static_cast<int>(value) << "(" << name << ")";
+    for (it = std::next(std::cbegin(map)); it != std::cend(map); std::advance(it, 1)) {
+        const auto& [name, value] = *it;
+        enumUsage << ", " << static_cast<int>(value) << "(" << name << ")";
     }
     enumUsage << " }";
 
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -105,27 +105,24 @@ NearObjectCli::CreateParser() noexcept
 namespace detail
 {
 /**
- * @brief Helper function to produce a mapping of enumeration strings to their
- * corresponding field names.
- *
- * This is used for CLI11 CheckedTransformer which validates string inputs and
- * maps them to their corresponding enumeration value.
+ * @brief Helper function to produce a mapping of enumeration field names to their
+ * corresponding values.
  *
  * @tparam EnumType The type of the enumeration.
- * @return std::map<std::string, EnumType>
+ * @return std::map<std::string, int>
  */
 template <typename EnumType>
 // clang-format off
 requires std::is_enum_v<EnumType>
-const std::unordered_map<std::string, EnumType>
+const std::unordered_map<std::string, int>
 CreateEnumerationStringMap() noexcept
 // clang-format on
 {
     const auto& reverseMap = magic_enum::enum_entries<EnumType>();
-    std::unordered_map<std::string, EnumType> map;
+    std::unordered_map<std::string, int> map;
 
     for (const auto& [enumValue, enumName] : reverseMap) {
-        map.emplace(enumName, enumValue);
+        map.emplace(enumName, static_cast<int>(enumValue));
     }
 
     return map;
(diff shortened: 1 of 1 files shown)


abeltrano:
Could you please attach a screenshot of the updated output?

tools/cli/NearObjectCli.cxx
```diff
@@ -105,27 +105,24 @@ NearObjectCli::CreateParser() noexcept
...
- * corresponding field names.
- *
- * This is used for CLI11 CheckedTransformer which validates string inputs and
- * maps them to their corresponding enumeration value.
+ * @brief Helper function to produce a mapping of enumeration field names to their
+ * corresponding values.
  *
  * @tparam EnumType The type of the enumeration.
- * @return std::map<std::string, EnumType>
+ * @return std::map<std::string, int>
```
abeltrano, line 112: Looks like this comment needs updating (should say `unordered_map` here instead of `map`).

abeltrano:
> ### Future Work
> By nature of std::unordered_map, there is no guarantee in the order in which the enum values are printed when using --help. This doesn't actually matter but it would be nice to have values printed in order (e.g. for --Channel: C5, C6, C8, ...). However, simply switching to std::map doesn't work because of situations where numbers are represented as strings (e.g. 10 would come before 5).

This is a good point and I'd been wondering about this myself. While it's nice to automate generating the help text using the enum names, the help text would be more clear if manually curated. So likely we should abandon this automated approach in the future and just write out the help text directly in a cleaner fashion.

I created #143 to track this.

tools/cli/NearObjectCli.cxx
```diff
@@ -152,9 +148,17 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo)
 {
     std::string optionName{ std::string("--").append(magic_enum::enum_type_name<EnumType>()) };
 
-    return app->add_option(std::move(optionName), assignTo)
-        ->transform(CLI::CheckedTransformer(CreateEnumerationStringMap<EnumType>(), CLI::ignore_case))
-        ->capture_default_str();
+    const auto& map = CreateEnumerationStringMap<EnumType>();
```
abeltrano, line 151: This is creating a reference to a temporary, so will need to drop that:

```suggestion
    const auto map = CreateEnumerationStringMap<EnumType>();
```

tools/cli/NearObjectCli.cxx
```diff
@@ -105,27 +105,24 @@ NearObjectCli::CreateParser() noexcept
...
 CreateEnumerationStringMap() noexcept
 // clang-format on
 {
     const auto& reverseMap = magic_enum::enum_entries<EnumType>();
-    std::unordered_map<std::string, EnumType> map;
+    std::unordered_map<std::string, int> map;
 
     for (const auto& [enumValue, enumName] : reverseMap) {
-        map.emplace(enumName, enumValue);
+        map.emplace(enumName, static_cast<int>(enumValue));
```
abeltrano, line 125: Why do the casting here, instead of where the map is used? Casting an enum value to an `int` will truncate data when `sizeof(std::underlying_type_t<EnumType>) > sizeof(int)`. I'd much rather see this function kept as-is and avoid this cast since it's generic, then just cast in the loop in `AddEnumOption()` if determined to be safe.

tools/cli/NearObjectCli.cxx
```diff
@@ -152,9 +148,17 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo)
...
-        ->transform(CLI::CheckedTransformer(CreateEnumerationStringMap<EnumType>(), CLI::ignore_case))
-        ->capture_default_str();
+    const auto& map = CreateEnumerationStringMap<EnumType>();
+    std::ostringstream enumUsage;
+
+    auto it = std::cbegin(map);
+    enumUsage << "value in { " << it->second << "(" << it->first << ")";
+    for (it = std::next(std::cbegin(map)); it != std::cend(map); ++it) {
+        enumUsage << ", " << it->second << "(" << it->first << ")";
+    }
```
abeltrano, line 158: Let's use a structured binding here for the map values to improve readability and comprehension. Someone coming upon this code won't know what `it->first` and `it->second` are, but a structured binding makes this super clear and sacrifices nothing:

```suggestion
    for (it = std::next(std::cbegin(map)); it != std::cend(map); ++it) {
        const auto& [name, value] = *it;
        enumUsage << ", " << value << "(" << name << ")";
    }
```

abeltrano (changes requested):

tools/cli/NearObjectCli.cxx
```diff
@@ -152,9 +148,17 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo)
...
 
-    return app->add_option(std::move(optionName), assignTo)
-        ->transform(CLI::CheckedTransformer(CreateEnumerationStringMap<EnumType>(), CLI::ignore_case))
-        ->capture_default_str();
+    const auto& map = CreateEnumerationStringMap<EnumType>();
+    std::ostringstream enumUsage;
+
+    auto it = std::cbegin(map);
+    enumUsage << "value in { " << it->second << "(" << it->first << ")";
+    for (it = std::next(std::cbegin(map)); it != std::cend(map); ++it) {
```
abeltrano, line 156: alternate here for consistency with function iterator access:

```suggestion
    for (it = std::next(std::cbegin(map)); it != std::cend(map); std::advance(it)) {
```

tools/cli/NearObjectCli.cxx
abeltrano: On second thought, can we just use that directly for the loop?

```suggestion
    for (const auto& [name, value] : map) {
        enumUsage << ", " << value << "(" << name << ")";
    }
```

corbin-phipps: I was using it directly for the loop at first, but then I ran into the fencepost problem with the comma separate list. I have made a change to match your original suggestion

corbin-phipps:
MicrosoftTeams-image

abeltrano:
> MicrosoftTeams-image

Cool, this looks nicer!

tools/cli/NearObjectCli.cxx
abeltrano: Why didn't the second suggestion (which is much cleaner) work?

abeltrano (approved):

tools/cli/NearObjectCli.cxx
corbin-phipps: That's what I was referring to with the fencepost problem. If I do "for (const auto& [name, value] : map)", that will iterate through the entire map, meaning the output would start with a ", ". So to prevent that, I print the first value before the loop, then have the loop iterate through the map starting with the second entry so that the commas make sense

abeltrano: Gotcha, I didn't read your comment and code carefully enough, thanks for the explanation. For future reference, a [`std::ranges::take_view`](https://en.cppreference.com/w/cpp/ranges/take_view) would also solve this problem.

PR #146, opened 2023-03-17
Add mandatory app config params to nocli

corbin-phipps:
### Type

- [ ] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR adds mandatory app config params to the nocli tool interface and ensures that each value is set.

### Technical Details

- Added new UwbMacAddress constructor that takes a mac address string and parses it, such as "12:34" for a short address
- Add NumberOfControlees and DeviceType mandatory params to UwbConfigurationData (future work includes adding these to UwbConfiguration as well)
- Adds logic to CLI tool to accept short or extended MAC addresses based on MacAddressMode (short is default)
- Adds logic to CLI tool to ensure that all mandatory params are set, displaying error otherwise

### Test Results

nocli has expected behavior. This includes only proceeding if all mandatory params are set and catching errors with improperly formatted mac address strings.

### Reviewer Focus

None.

### Future Work

Not all of the mandatory app config params are part of the UwbConfiguration struct (no ParameterTag exists for them), so those values aren't printed to the CLI output. Fixing this design to not rely on ParameterTag will help resolve this issue.

Not all params in UwbConfigurationData have options in the CLI, such as MacAddressMode. These should be added at some point.

### Checklist


tools/cli/NearObjectCli.cxx
```diff
@@ -227,12 +227,25 @@ NearObjectCli::AddSubcommandUwbRangeStart(CLI::App* parent)
 
     // TODO is there a way to put all the enums into a list of [optionName, optionDestination, optionMap] so we don't have to create the initializer list each time
 
-    // enumerations
+    // List mandatory params first
+    // TODO: --help should output that these params are mandatory
     detail::AddEnumOption(rangeStartApp, uwbConfig.deviceRole);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.multiNodeMode);
+    rangeStartApp->add_option("--NumberOfControlees", uwbConfig.numberOfControlees, "1 <= N <= 8")->capture_default_str(); // TODO: Input validation
```
forwardpointer, line 234: You can indicate that a param is mandatory on the help text by doing something like
```
rangeStartApp->add_option("--NumberOfControlees", uwbConfig.numberOfControlees, "1 <= N <= 8   This parameter is mandatory for start ranging")->capture_default_str();
```

abeltrano: I'm wondering if we care to specify this at all. If we allow the user/caller to specify N controlees, then we'll know exactly how many controlees there are. My understanding is that if you specify `N` here, then `N` controlee addresses need to be specified as well, which would make this redundant. Thoughts?

corbin-phipps: Oh yeah that's true. It'll take an extra step to do something similar for the enums using AddEnumOption, but I can make it work

lib/uwb/UwbMacAddress.cxx
```diff
@@ -25,6 +27,43 @@ UwbMacAddress::UwbMacAddress(const UwbMacAddress& other) :
     InitializeView();
 }
 
+UwbMacAddress::UwbMacAddress(std::string addressString, UwbMacAddressType addressType)
```
abeltrano, line 30: Can you please add tests for this [here](https://github.com/microsoft/nearobject-framework/blob/develop/tests/unit/uwb/TestUwbMacAddress.cxx)?

lib/uwb/UwbMacAddress.cxx
```diff
@@ -25,6 +27,43 @@ UwbMacAddress::UwbMacAddress(const UwbMacAddress& other) :
...
+
+        if (addressType == UwbMacAddressType::Short) {
+            *this = UwbMacAddress(shortAddress);
+        } else {
+            *this = UwbMacAddress(extendedAddress);
+        }
+    }
+    catch (const std::exception& e)
+    {
+        throw std::invalid_argument("Invalid mac address string: " + addressString);
```
abeltrano, line 63: We're trying to avoid throwing in this class. You can avoid that if you make this constructor private and then wrap its use in a static factory function which returns a `std::optional<::uwb::UwbMacAddress>` instead, eg.

```
static
std::optional<::uwb::UwbMacAddress> FromString(const std::string addressString, UwbMacAddressType macAddressType)
{
    try {
        return UwbMacAddress(addressString, macAddressType);
    } catch (...) {
        return std::nullopt;
    }
}
```

This way the caller doesn't have to wrap creating a mac address in a try/catch block.

forwardpointer (dismissed):

lib/uwb/include/uwb/UwbMacAddress.hxx
```diff
@@ -361,6 +361,15 @@ public:
      */
     UwbMacAddress(const UwbMacAddress& other);
 
+    /**
+     * @brief Construct a new UwbMacAddress object given an input string of
+     * semicolon-delimited bytes.
```
abeltrano, line 366: It'd be nice if the delimiter was configurable and/or optional. For example, if someone wanted to specify `0A0B` on the command line, I don't see why we shouldn't allow it.

No need to hold up the review for this, but perhaps adding a TODO somewhere here indicating this would be good.

abeltrano (changes requested):

tools/cli/NearObjectCli.cxx
corbin-phipps: I think we still need this, only for the case when the user is configuring a controlee device in a multicast session. In this scenario, DeviceType will be Controlee, DeviceMacAddress (from UCI perspective) will be this particular device's mac address, and DstMacAddress (from UCI perspective) will be the Controller's mac address. So the user isn't going to input the mac addresses of ALL controlees, but the number of controlees is still needed.

However, now that I type this out, I'm realizing that I was originally intending to allow the user of the tool to specify the Controller and N Controlees, which conflicts with what I just said. Do you think there is value in the user of this tool to have to supply all of the information (meaning specify the Controller and all N Controlees), or should the user only care about their own device (meaning specify the Controller and itself, plus number of controlees)?

lib/uwb/UwbMacAddress.cxx
corbin-phipps: Yes, I added tests locally and it helped me find a bug.. so I will push new changes once fixed!

lib/uwb/UwbMacAddress.cxx
corbin-phipps: I made this change along with some others to make all error handling more graceful. It's much better now!

tools/cli/NearObjectCli.cxx
abeltrano: > Do you think there is value in the user of this tool to have to supply all of the information (meaning specify the Controller and all N Controlees), or should the user only care about their own device (meaning specify the Controller and itself, plus number of controlees)?

I think your interpretation is correct as the `NUMBER_OF_CONTROLEES` shows up in CSML `UWB_CONFIGURATION` where `DST_MAC_ADDRESS` isn't described at all (only the controlee short mac and controller mac, both singular, show up):

image
image

meaning the number of controlees can't be tied to `DST_MAC_ADDRESSES` and so it makes sense that there's a separate application configuration parameter for it.

Consequently, I think we have it encoded wrong in the CLI. We should allow specifying the "self" mac address (`DeviceMacAddress`) which is only ever singular, and then the "destination" mac addresses, of which there can be many. For the self mac address, we should also allow an option to generate a random address (`UwbMacAddress::Random()` supports this). I also agree with you that specifying all the controlee addresses should only be needed when the host is a controller:

| Device Type  | Device Mac Address | Destination Mac Addresses |
| --- |  --- | --- |
| Controller | Own Address  | 1<= N <= 8 Controlee Addresses |
| Controlee | Own Address  | Controller Address |

tools/cli/NearObjectCli.cxx
```diff
@@ -157,6 +158,10 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo)
     }
     enumUsage << " }";
 
+    if (isMandatory) {
+        enumUsage << " This parameter is mandatory for a ranging session";
+    }
```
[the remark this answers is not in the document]
corbin-phipps: I don't know of a way to mark the arg as mandatory, hence the added message to the --help usage. Do you mean simply NOT using std::optional for these?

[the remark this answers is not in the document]
abeltrano: I think the `CLI11:App` type has a `required()` function to do this, so you can append this to the end of the `add_option()` call below.

tools/cli/NearObjectCli.cxx
corbin-phipps: I agree about the encoding, so I will change the CLI to have an option for DeviceMacAddress (referring to "self" mac address), as well as DestinationMacAddresses (either controller is host is a controlee, or all controlees if host is controller). Unfortunately, I would have to list all the options all at once due to the nature of command line arguments. Ideally, we would first see what the DeviceType is (controller/controlee), then display the other arguments depending on that value, but I don't think that's possible

tools/cli/NearObjectCli.cxx
```diff
@@ -227,22 +232,33 @@ NearObjectCli::AddSubcommandUwbRangeStart(CLI::App* parent)
...
+    detail::AddEnumOption(rangeStartApp, uwbConfig.rangingDirection, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.rangingMeasurementReportMode, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.stsConfiguration, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.rangingTimeStruct, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.schedulingMode, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.channel, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.rframeConfig, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.convolutionalCodeConstraintLength, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.prfMode, false);
+    detail::AddEnumOption(rangeStartApp, uwbConfig.macAddressFcsType, false);
```
abeltrano, line 259: Instead of specifying `false` for the majority of these, why not provide a default `true` value for the `isMandatory` argument. This way, the arg is only specified for the exceptional cases and makes the code less verbose.

tools/cli/NearObjectCli.cxx
```diff
@@ -269,6 +285,47 @@ NearObjectCli::AddSubcommandUwbRangeStart(CLI::App* parent)
...
+        } else { // uwb::protocol::fira::DeviceType::Controlee
+            controllerMacAddress = uwb::UwbMacAddress::FromString(m_cliData->destinationMacAddress, macAddressType);
+            controleeMacAddress = uwb::UwbMacAddress::FromString(m_cliData->deviceMacAddress, macAddressType);
+        }
+
+        if (!controllerMacAddress.has_value()) {
+            std::cerr << "Invalid ControllerMacAddress" << std::endl;
+        } else {
+            m_cliData->uwbConfiguration.controllerMacAddress = controllerMacAddress.value();
+            // TEST
```
abeltrano, line 317: Does this comment and/or associated code need to be removed, or are they here intentionally?

tests/unit/uwb/TestUwbMacAddress.cxx
```diff
@@ -321,4 +321,73 @@ TEST_CASE("uwb address can be used with stream operators", "[basic][io]")
...
+    };
+
+    static const auto ExtendedUwbMacAddressStrings = {
+        "11:22:33:44:55:66:77:88",
+        "AA:BB:CC:DD:DD:CC:BB:AA",
+        "aA:bB:cC:dD:Dd:Cc:Bb:Aa",
+        "A1:B2:C3:D4:1a:2b:3c:4d"
+    };
+
+    static const auto ExtendedUwbMacAddressStringsInvalid = {
```
abeltrano, line 352: Should probably also add the empty string as a test somewhere.

lib/uwb/UwbMacAddress.cxx
```diff
@@ -100,15 +132,34 @@ UwbMacAddress::Random(UwbMacAddressType type)
     }
 }
 
+/* static */
+std::optional<UwbMacAddress>
+UwbMacAddress::FromString(const std::string& addressString, UwbMacAddressType addressType)
+{
+    std::regex shortRegex("^([0-9A-Fa-f]{2}:){1}([0-9A-Fa-f]{2})$");
+    std::regex extendedRegex("^(([0-9A-Fa-f]{2}:){7}[0-9A-Fa-f]{2})$");
```
abeltrano, line 140: Can these be `const`?

abeltrano (approved):

abeltrano:
Looks like the new constructor taking `std::string` got removed in this merge commit:
https://github.com/microsoft/nearobject-framework/pull/146/commits/b78d3aefb9042a9fc68d7328288dae5c43486ff7

lib/uwb/UwbMacAddress.cxx
```diff
@@ -39,6 +41,36 @@ UwbMacAddress::GetValueShort() const
     return (static_cast<uint16_t>(m_view[1]) << 8U) | m_view[0];
 }
 
+UwbMacAddress::UwbMacAddress(const std::string& addressString, UwbMacAddressType addressType)
```
corbin-phipps, line 44: I was unable to make this one a constexpr like the other constructors due to std::stringstream. Perhaps there is a way around it that I don't know about?

forwardpointer (approved):

PR #149, opened 2023-03-17
Add UCI app config params to UwbConfiguration

abeltrano:
This PR draft considers whether we should add the UCI APP_CONFIG_PARAM types to the existing `UwbConfiguration` object. It's a draft to solicit opinions on this, vs having a separate type to represent them. That can be found in #148.


abeltrano:
Closing in favor of #148.

PR #154, opened 2023-03-21
Configure impl

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [x] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR changes the interface of UwbSession::Configure to take in the sessionId and a vector of UwbApplicationConfigurationParameter because those are what UCI defines as inputs to the driver for the IOCTL.

This PR also fixes some minor documentation issues with the UwbDeviceConnector

Also the UwbConfiguration::uwbConfigurationAvailable boolean was deemed unnecessary because you can just check the output of UwbConfiguration::GetValueMap(), and so it was removed.

### Future Work

We need to implement the translation function between the OOB app config params to the UCI config params. I have the function declared as 

```
std::vector<UwbApplicationConfigurationParameter>
uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& config, DeviceType deviceType)
```
and it is declared in lib/uwb/protocols/fira/UwbSession.hxx

### Checklist

### DIFF
--- a/lib/uwb/UwbSession.cxx
+++ b/lib/uwb/UwbSession.cxx
@@ -6,8 +6,10 @@
 
 #include <uwb/UwbSession.hxx>
 #include <uwb/UwbSessionEventCallbacks.hxx>
+#include <uwb/protocols/fira/UwbException.hxx>
 
 using namespace uwb;
+using namespace uwb::protocol::fira;
 
 UwbSession::UwbSession(std::weak_ptr<UwbSessionEventCallbacks> callbacks, uwb::protocol::fira::DeviceType deviceType) :
     m_uwbMacAddressSelf(UwbMacAddress::Random<UwbMacAddressType::Extended>()),
@@ -45,8 +47,16 @@ UwbSession::AddPeer(UwbMacAddress peerMacAddress)
 void
 UwbSession::Configure(const uint32_t sessionId, const std::vector<protocol::fira::UwbApplicationConfigurationParameter> configParams)
 {
-    PLOG_VERBOSE << "configure";
-    ConfigureImpl(sessionId, configParams);
+    PLOG_VERBOSE << "configure session with id " << m_sessionId;
+    try {
+        ConfigureImpl(sessionId, configParams);
+    } catch (UwbException& uwbException) {
+        PLOG_ERROR << "error configuring session with id " << m_sessionId << ", status=" << ToString(uwbException.Status);
+        throw uwbException;
+    } catch (std::exception& e) {
+        PLOG_ERROR << "error configuring session with id " << m_sessionId << ", unexpected exception status=" << e.what();
+        throw e;
+    }
 }
 
 void
--- a/tools/cli/NearObjectCliHandler.cxx
+++ b/tools/cli/NearObjectCliHandler.cxx
@@ -1,4 +1,6 @@
 
+#include <plog/Log.h>
+
 #include <nearobject/cli/NearObjectCliHandler.hxx>
 #include <nearobject/cli/NearObjectCliUwbSessionEventCallbacks.hxx>
 #include <uwb/UwbDevice.hxx>
@@ -6,8 +8,10 @@
 #include <uwb/protocols/fira/UwbOobConversions.hxx>
 
 #include <plog/Log.h>
+#include <uwb/protocols/fira/UwbException.hxx>
 
 using namespace nearobject::cli;
+using namespace uwb::protocol::fira;
 
 std::shared_ptr<uwb::UwbDevice>
 NearObjectCliHandler::ResolveUwbDevice(const nearobject::cli::NearObjectCliData& /*cliData */) noexcept
@@ -19,34 +23,44 @@ NearObjectCliHandler::ResolveUwbDevice(const nearobject::cli::NearObjectCliData&
 
 void
 NearObjectCliHandler::HandleStartRanging(std::shared_ptr<uwb::UwbDevice> uwbDevice, uwb::protocol::fira::UwbSessionData& sessionData) noexcept
-{
+try {
     auto callbacks = std::make_shared<nearobject::cli::NearObjectCliUwbSessionEventCallbacks>();
     auto session = uwbDevice->CreateSession(callbacks);
     session->Configure(sessionData.sessionId, uwb::protocol::fira::GetUciConfigParams(sessionData.uwbConfiguration, session->GetDeviceType()));
     session->StartRanging();
+} catch (...) {
+    PLOG_ERROR << "failed to start ranging";
 }
 
 void
 NearObjectCliHandler::HandleStopRanging() noexcept
-{
+try {
     // TODO
+} catch (...) {
+    PLOG_ERROR << "failed to stop ranging";
 }
 
 void
 NearObjectCliHandler::HandleMonitorMode() noexcept
-{
+try {
     // TODO
+} catch (...) {
+    PLOG_ERROR << "failed to initiate monitor mode";
 }
 
 void
 NearObjectCliHandler::HandleDeviceReset(std::shared_ptr<uwb::UwbDevice> uwbDevice) noexcept
-{
+try {
     uwbDevice->Reset();
+} catch (...) {
+    PLOG_ERROR << "failed to reset uwb device";
 }
 
 void
 NearObjectCliHandler::HandleGetDeviceInfo(std::shared_ptr<uwb::UwbDevice> uwbDevice) noexcept
-{
+try {
     auto deviceInfo = uwbDevice->GetDeviceInformation();
     std::cout << deviceInfo.ToString() << std::endl;
+} catch (...) {
+    PLOG_ERROR << "failed to obtain device information";
 }
--- a/lib/uwb/include/uwb/protocols/fira/UwbException.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/UwbException.hxx
@@ -2,13 +2,15 @@
 #ifndef UWB_EXCEPTION_HXX
 #define UWB_EXCEPTION_HXX
 
+#include <exception>
 #include <utility>
 
 #include <uwb/protocols/fira/FiraDevice.hxx>
 
 namespace uwb::protocol::fira
 {
-struct UwbException
+struct UwbException :
+    public std::exception
 {
     explicit UwbException(UwbStatus status) :
         Status(std::move(status))
--- a/windows/devices/uwb/CMakeLists.txt
+++ b/windows/devices/uwb/CMakeLists.txt
@@ -67,6 +67,7 @@ target_include_directories(windev-uwb
 
 target_link_libraries(windev-uwb
     PRIVATE
+        magic_enum::magic_enum
         windev-util
         windev-uwb-cx-adapter
     PUBLIC
--- a/windows/devices/uwb/UwbSession.cxx
+++ b/windows/devices/uwb/UwbSession.cxx
@@ -4,9 +4,11 @@
 #include <memory>
 #include <numeric>
 
+#include <magic_enum.hpp>
 #include <plog/Log.h>
 #include <wil/result.h>
 
+#include <uwb/protocols/fira/UwbException.hxx>
 #include <windows/devices/uwb/UwbCxDdiLrp.hxx>
 #include <windows/devices/uwb/UwbSession.hxx>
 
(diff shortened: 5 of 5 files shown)


windows/devices/uwb/include/windows/devices/uwb/UwbCxAdapter.hxx
```diff
@@ -11,7 +11,7 @@
 namespace windows::devices::uwb
 {
 /**
- * @brief TODO
+ * @brief Generate neutral type UwbSetAppConfigParameters from UwbSessionData TODO should this function be located in UwbCxAdapterDdiLrp.hxx?
```
forwardpointer, line 14: this file needs to be removed, not used anywhere

abeltrano: Yes, I kept it so that we can use the code from it to implement a few functions for the new type. So this is intentional for now.

abeltrano (approved):

lib/uwb/UwbSession.cxx
```diff
@@ -36,10 +36,10 @@ UwbSession::AddPeer(UwbMacAddress peerMacAddress)
 }
 
 void
-UwbSession::Configure(const uwb::protocol::fira::UwbSessionData& uwbSessionData)
+UwbSession::Configure(uint32_t sessionId, const std::vector<protocol::fira::UwbApplicationConfigurationParameter> configParams)
```
corbin-phipps, line 39: Can sessionId be const as well?

lib/uwb/protocols/fira/UwbConfiguration.cxx
```diff
@@ -18,6 +18,116 @@ const std::unordered_set<ResultReportConfiguration> UwbConfiguration::ResultRepo
     ResultReportConfiguration::AoAAzimuthReport
 };
 
+/**
+ * @brief map of uci tag to a function of how to generate that uci tag given UwbConfiguration
+ *
+ */
+const std::unordered_map<uwb::protocol::fira::UwbApplicationConfigurationParameterType, std::function<std::optional<uwb::protocol::fira::UwbApplicationConfigurationParameterValue>(UwbConfiguration*)>> uciGenerators{
```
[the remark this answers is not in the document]
forwardpointer: Yea I'm actually considering if it might be more appropriate to move the oob2uci translation to a separate header, one that only UwbSession includes, so that way UwbConfiguration doesn't have to know about the uci type and vice versa

lib/uwb/protocols/fira/UwbConfiguration.cxx
```diff
@@ -18,6 +18,116 @@ const std::unordered_set<ResultReportConfiguration> UwbConfiguration::ResultRepo
...
+     } },
+    { uwb::protocol::fira::UwbApplicationConfigurationParameterType::HoppingMode, [](UwbConfiguration* config) -> std::optional<uwb::protocol::fira::UwbApplicationConfigurationParameterValue> {
+         return config->GetHoppingMode();
+     } },
+    // { uwb::protocol::fira::UwbApplicationConfigurationParameterType::ResultReportConfig, [](UwbConfiguration* config) -> std::optional<uwb::protocol::fira::UwbApplicationConfigurationParameterValue> {
+    //      return config->GetResultReportConfigurations();
+    //  } },
+
+    // params with different names
+    // { uwb::protocol::fira::UwbApplicationConfigurationParameterType::RangingRoundUsage, [](UwbConfiguration* config) -> std::optional<uwb::protocol::fira::UwbApplicationConfigurationParameterValue> {
```
corbin-phipps: Why is this one commented out? I believe that RangingRoundUsage (UCI) and RangingMethod (CSML) correspond to the same exact thing, so this should be fine to include here

tools/cli/NearObjectCliHandler.cxx
```diff
@@ -19,8 +21,12 @@ NearObjectCliHandler::HandleStartRanging(std::shared_ptr<uwb::UwbDevice> uwbDevi
 {
     auto callbacks = std::make_shared<nearobject::cli::NearObjectCliUwbSessionEventCallbacks>();
     auto session = uwbDevice->CreateSession(callbacks);
-    session->Configure(sessionData);
-    session->StartRanging();
+    if (not sessionData.uwbConfigurationAvailable) {
+        PLOG_WARNING << "did not start ranging due to missing configuration";
```
corbin-phipps, line 25: I don't think this is correct. If the user of the nocli tool is starting a ranging session, that means they are starting a session from scrach and will be setting the config params as CLI args. So, they will see this warning each time and the session will never be configured or ran, which doesn't make sense. Unless I'm misunderstanding the if statement here

corbin-phipps (changes requested):

lib/uwb/protocols/fira/UwbConfiguration.cxx
forwardpointer: There's an issue with the neutral and DDI types, so I commented them out for now.

tools/cli/NearObjectCliHandler.cxx
forwardpointer: My understanding is that HandleStartRanging is called when the user enters the range start command via CLI. This means they need to input the config params to start ranging, which I think is correct.

forwardpointer: The 
```
nocli uwb range start [--options]
```
command has the user enter in the config params via the --options

forwardpointer: Actually you raised a good point. Because UwbConfiguration right now is a bunch of std::optional this means we don't actually need a boolean to keep track of whether there is a UwbConfiguration. And it looks like that boolean is really being used anywhere either. So I'll remove this unnecessary check.

corbin-phipps (dismissed):

lib/uwb/protocols/fira/UwbConfiguration.cxx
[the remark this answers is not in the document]
forwardpointer: I just pushed the change that moves the translation function to UwbSession.hxx instead of being tied to UwbConfiguration

corbin-phipps (approved):

lib/uwb/protocols/fira/UwbConfiguration.cxx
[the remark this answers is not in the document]
forwardpointer: It's now in its own header, located in UwbOobConversion.hxx

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -0,0 +1,126 @@
...
+         return std::nullopt;
+     } },
+    { UwbApplicationConfigurationParameterType::DestinationMacAddresses, [](const UwbConfiguration& config, DeviceType deviceType) -> std::optional<UwbApplicationConfigurationParameterValue> {
+         auto mode = config.GetMacAddressMode();
+         if (not mode) {
+             return std::nullopt;
+         }
+         if (deviceType == DeviceType::Controller) {
+             if (mode == uwb::UwbMacAddressType::Short) {
+                 return config.GetControleeShortMacAddress(); // TODO this should reflect the possibility for multiple peers
```
abeltrano, line 100: This might not be necessary. If our current understanding is accurate, then we'll expect many `UwbSessionData` (with `UwbConfiguration` in them) to be provided via OOB, so we'd call `UwbSession::Configure` for each of them after converting them with this code, and each of them should only contain a single address for the controllee. Eg.

```
DeviceType deviceType = DeviceType::Controller; // from OOB "profile"
UwbSession uwbSession = uwbDevice.CreateSession(deviceType);
foreach (UwbSessionData from OOB /* = N */) {
    auto applicationConfigurationParameters = GetUciConfigParams(UwbSessionData.uwbConfiguration, deviceType);
    uwbSession.Configure(uwbSessionData.sessionId, std::move(applicationConfigurationParameters));
}
// uwbSession is now configured for 'N' sets of parameters for 'N' controlees
```

I think that when the host is a controller, it will set a lot of the application configuration parameters itself, whereas when the host is a controlee, it gets most of the parameters from the controller, and there will likely only be one `UwbSessionData` from OOB.

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -0,0 +1,126 @@
...
+             } else {
+                 // TODO what do we do here
+                 return std::nullopt;
+             }
+         } else {
+             return config.GetControllerMacAddress();
+         }
+     } },
+    // TODO figure out the rest of the uci params
```
abeltrano, line 109: Are there actually any more to do now?

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
```diff
@@ -549,7 +549,7 @@ using UwbApplicationConfigurationParameterValue = std::variant<
     RangingRoundUsage, // RANGING_ROUND_USAGE, tag0x01, size 1
     RangingMode, // RANGING_TIME_STRUCT, tag 0x1A, size 1
     RangingRoundControl, // RANGING_ROUND_CONTROL, tag 0x0C, size 1,
-    ResultReportConfiguration, // RESULT_REPORT_CONFIG, tag 0x2E, size 1
+    ResultReportConfiguration, // RESULT_REPORT_CONFIG, tag 0x2E, size 1 // TODO this should really be a set
```
abeltrano, line 552: Might as well fix this now so we can close this chapter.

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -24,16 +24,24 @@ class UwbSession
...
      */
     virtual ~UwbSession() = default;
 
+    /**
+     * @brief Get the Device Type associated with the host of this UwbSession
+     * 
+     * @return uwb::protocol::fira::DeviceType 
+     */
+    uwb::protocol::fira::DeviceType
+    GetDeviceType();
```
abeltrano, line 43: This should be `const except`:
```suggestion
    GetDeviceType() const noexcept;
```

lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
```diff
@@ -0,0 +1,20 @@
...
+
+namespace uwb::protocol::fira
+{
+/**
+ * @brief Converts the config params given by OOB to config params that UCI needs
+ *
+ * @return std::vector<UwbApplicationConfigurationParameter>
+ */
+std::vector<UwbApplicationConfigurationParameter>
+GetUciConfigParams(const UwbConfiguration& uwbConfiguration, DeviceType deviceType);
```
abeltrano, line 16: This is great here.

lib/uwb/include/uwb/protocols/fira/UwbSessionData.hxx
```diff
@@ -94,8 +93,7 @@ struct hash<uwb::protocol::fira::UwbSessionData>
             uwbSessionData.subSessionId,
             uwbSessionData.uwbConfiguration,
             uwbSessionData.staticRangingInfo,
-            uwbSessionData.secureRangingInfo,
-            uwbSessionData.uwbConfigurationAvailable);
```
abeltrano, line 98: Thank you for identifying this was unused and removing it!

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -0,0 +1,126 @@
+
+#include <uwb/protocols/fira/UwbOobConversions.hxx>
+
+using namespace uwb::protocol::fira;
+/**
+ * @brief map of uci tag to a function of how to generate that uci tag given UwbConfiguration
```
abeltrano, line 6: Can we say "parameter type" instead of tag to avoid confusion with the TLV tags used in the OOB structures?
```suggestion
 * @brief map of uci parameter type to a function of how to generate that parameter value given UwbConfiguration
```
Same goes for other places here where `tag` is used in the code.

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
forwardpointer: I would prefer to do this in another PR (keeps this one small)

lib/uwb/protocols/fira/UwbOobConversions.cxx
abeltrano: Ok, so we'll defer these to later.

abeltrano (approved):

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
abeltrano: 👍

abeltrano (approved):

Issue #126, opened 2023-03-08
Should there be two interfaces that UwbDeviceConnector implements, one that only UwbSession interacts with, and one that only UwbDevice interacts with?


forwardpointer:
Currently lots of things in UwbDeviceConnector are public, but also only pertain to only UwbSession or UwbDevice.

PR #165, opened 2023-03-23
change interface to IUwbSessionDdiConnector and IUwbDeviceDdiConnector

forwardpointer:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [X] Infrastructure

### Goals

Address #126 

### Technical Details

UwbDeviceConnector implements two interfaces
 * IUwbDeviceDdiConnector contains exactly those functions that talk with the DDI that UwbDevice is interested in
 * IUwbSessionDdiConnector does the same for UwbSession

I moved the RegisteredCallbackToken to be forward declared in UwbRegisteredCallbacks, but it's definition is in UwbDeviceConnector.cxx
 * I feel it makes more sense to have its forward declaration in the same file as the callback structs
 
### Test Results

None yet

### Reviewer Focus
Does this split in interfaces make sense? Technically, UwbDevice.cxx _is_ still able to access all of the capabilities of UwbDeviceConnector by simply casting its m_uwbDeviceConnector but I think the extra step of casting should signal to any user that they need to use caution.

### Checklist


windows/devices/uwb/include/windows/devices/uwb/IUwbDeviceDdi.hxx
```diff
@@ -32,46 +37,50 @@ struct IUwbDeviceDdi
...
+     * @return false If listening for notifications could not be started.
+     */
+    virtual bool
+    NotificationListenerStart() = 0;
+
+    /**
+     * @brief Stop listening for notifications.
+     */
+    virtual void
+    NotificationListenerStop() = 0;
```
abeltrano, line 81: These should eventually be removed. The implementation should figure out when to start and stop listening based on the presence of callbacks. If it's easier, feel free to do that in a separate PR.

windows/devices/uwb/include/windows/devices/uwb/IUwbDeviceDdi.hxx
```diff
@@ -32,46 +37,50 @@ struct IUwbDeviceDdi
...
+
+struct IUwbDeviceDdiConnector : public IUwbDeviceDdi
+{
+    /**
+     * @brief Get the name of this device.
+     *
+     * @return const std::string&
+     */
+    virtual const std::string&
+    DeviceName() const noexcept = 0;
```
abeltrano, line 53: This shouldn't be in the interface since it doesn't map to a DDI function. It can easily stay in `UwbDeviceConnector` though.

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -41,13 +37,15 @@ public:
...
+    // IUwbDeviceDdiConnector
     /**
      * @brief Get the name of this device.
      *
      * @return const std::string&
      */
-    const std::string&
-    DeviceName() const noexcept;
+    virtual const std::string&
+    DeviceName() const noexcept override;
```
abeltrano, line 48: As mentioned earlier, let's drop this from `IUwbDeviceDdiConnector` and just expose it directly form this class.
```suggestion
    const std::string&
    DeviceName() const noexcept;
```

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -16,17 +16,13 @@
...
 
 namespace windows::devices::uwb
 {
-/**
- * @brief Opaque class forward declaration to help with the deregistration
- *
- */
-class RegisteredCallbackToken;
-
 class UwbDeviceConnector :
```
abeltrano, line 23: This is a connector for both device and session based DDI functions, so let's rename to `UwbConnector`.
```suggestion
class UwbConnector :
```

windows/devices/uwb/UwbSession.cxx
```diff
@@ -15,7 +15,7 @@
 using namespace windows::devices::uwb;
 using namespace ::uwb::protocol::fira;
 
-UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<::uwb::UwbSessionEventCallbacks> callbacks, std::shared_ptr<UwbDeviceConnector> uwbDeviceConnector, ::uwb::protocol::fira::DeviceType deviceType) :
+UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<::uwb::UwbSessionEventCallbacks> callbacks, std::shared_ptr<IUwbSessionDdiConnector> uwbDeviceConnector, ::uwb::protocol::fira::DeviceType deviceType) :
```
abeltrano, line 18: As discussed offline, we'll remove the connector from here. However, I think we should complete this PR before the discussed changes should be made, so we can leave this here for now. This comment is just to document the future plan, so you can resolve it without taking any action here.

abeltrano (approved):

windows/devices/uwb/include/windows/devices/uwb/IUwbDeviceDdi.hxx
forwardpointer: I moved it into UwbDeviceConnector but I'm unsure how users would call it since they are supposed to either talk only with IUwbDeviceDdiConnector or IUwbSessionDdiConnector

abeltrano: If moving it didn't break anything, that might be reason that it's not even needed. So, I wouldn't worry about it for now. Also, the thing that creates it will inherently know it's type, so if really necessary, it can do a `dynamic_cast` on the underlying pointer to get the real type instance.

abeltrano (approved):

corbin-phipps (approved):

PR #167, opened 2023-03-24
Fix bug where sessionId is not found after session has been initialized

forwardpointer:
### Type

- [X] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [X] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [X] Infrastructure

### Goals
Bug description:
CreateSession initializes the session with a default sessionId and inserts the session object into UwbDevice::m_sessions map. Then the sessionId is set after Configure has been called on it, but we never update the UwbDevice::m_sessions map with the new mapping

This PR makes the following changes to fix:
* provide the requested sessionId in the CreateSession command, which would mean the session would be inserted properly into the m_sessions map. 
* the Configure command would just take in the appConfigParams

### Test Results

None yet

### Checklist

### DIFF
--- a/lib/uwb/UwbSession.cxx
+++ b/lib/uwb/UwbSession.cxx
@@ -11,11 +11,11 @@
 using namespace uwb;
 using namespace uwb::protocol::fira;
 
-UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<UwbSessionEventCallbacks> callbacks, uwb::protocol::fira::DeviceType deviceType) :
-    m_sessionId(sessionId),
+UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<UwbSessionEventCallbacks> callbacks, DeviceType deviceType) :
     m_uwbMacAddressSelf(UwbMacAddress::Random<UwbMacAddressType::Extended>()),
     m_callbacks(std::move(callbacks)),
-    m_deviceType{ deviceType }
+    m_deviceType{ deviceType },
+    m_sessionId(sessionId)
 {}
 
 uwb::protocol::fira::DeviceType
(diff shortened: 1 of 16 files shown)


lib/uwb/UwbSession.cxx
```diff
@@ -11,7 +11,8 @@
 using namespace uwb;
 using namespace uwb::protocol::fira;
 
-UwbSession::UwbSession(std::weak_ptr<UwbSessionEventCallbacks> callbacks, uwb::protocol::fira::DeviceType deviceType) :
+UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<UwbSessionEventCallbacks> callbacks, uwb::protocol::fira::DeviceType deviceType) :
```
abeltrano, line 14: Should remove the namespace prefix in the source file since we have a `using` directive above to take care of it.
```suggestion
UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<UwbSessionEventCallbacks> callbacks, DeviceType deviceType) :
```

lib/uwb/UwbSession.cxx
```diff
@@ -11,7 +11,8 @@
...
+    m_sessionId(sessionId),
     m_uwbMacAddressSelf(UwbMacAddress::Random<UwbMacAddressType::Extended>()),
     m_callbacks(std::move(callbacks)),
     m_deviceType{ deviceType }
```
abeltrano, line 18: Can you re-order these so their order here matches their declaration order? Regardless of the order they appear here, they will always be initialized in the order they're declared, so if they appear out-of-order here, someone might think they're initialized in that order.

I think just `m_deviceType` needs to go first and all else look in order.

```suggestion
    m_deviceType{ deviceType },
    m_sessionId(sessionId),
    m_uwbMacAddressSelf(UwbMacAddress::Random<UwbMacAddressType::Extended>()),
    m_callbacks(std::move(callbacks))

```

abeltrano (changes requested):
Just a few small changes, but this is great 👍

PR #169, opened 2023-03-24
Separate driver and service testing functionality in nocli

corbin-phipps:
### Type

- [x] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR separates the driver testing functionality (under sub-command 'uwb') from service testing functionality (under new sub-command 'service'). This is necessary for configuring ranging sessions, because users testing drivers will need to set UCI app config params whereas users testing service code will need to set OOB params.

### Technical Details

- Created new top-level sub-command calls 'service'
- Restored original 'range start' code and moved under 'service' (nocli.exe service range start)
- Modified 'range start' code for driver testing and moved under 'uwb' (nocli.exe uwb range start)
- Added structures under NearObjectCliData.hxx for setting mandatory app config params for driver testing
- Added support for unordered_set<UwbMacAddress> in a conversion function in UwbCxAdapterDdiLrp.cxx
- Fixed implementation of UwbDeviceConnector::SetApplicationConfigurationParameters to use new wrapper for UWB_SET_APP_CONFIG_PARAMS

### Test Results

Ran nocli tool for driver testing ('uwb' sub-command) with newest TestClient driver and can successfully invoke necessary IOCTLs for starting a ranging session.

### Reviewer Focus

Note: The current code in HandleDriverStartRanging tries to get the app config params after configuring a session. Because the TestClient driver does not support this (yet!), I had to comment these few lines out in order for nocli to work.

### Future Work

- Need to ensure that stopping the ranging session will trigger StopRanging and SessionDeinit.
- Add rest of UCI app config params support in NearObjectCli.cxx. Currently, only the mandatory app config params are supported
- The code for 'uwb range start' could be improved. For example, the current code manually pushes each of the mandatory app config params into a std::vector<UwbApplicationConfigurationParameter>
- Multicast support (more than one destination MAC address / controlee).

### Checklist

### DIFF
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -404,11 +404,8 @@ NearObjectCli::AddSubcommandUwbRangeStart(CLI::App* parent)
             if (m_cliData->appConfigParamsData.numberOfControlees != 1) {
                 std::cerr << "Only 1 controlee expected in Unicast mode" << std::endl;
             }
-        } else {
-            if (m_cliData->appConfigParamsData.numberOfControlees < 1 || m_cliData->appConfigParamsData.numberOfControlees > MaxNumberOfControlees) {
-                std::cerr << "Invalid number of controlees. Must be 1 <= N <= 8" << std::endl;
-            }
         }
+
         // Set MAC addresses
         const auto macAddressType = m_cliData->appConfigParamsData.macAddressMode == uwb::UwbMacAddressType::Extended ? uwb::UwbMacAddressType::Extended : uwb::UwbMacAddressType::Short;
 
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -93,27 +93,51 @@ NearObjectCli::GetUwbApp() noexcept
 }
 
 CLI::App&
-NearObjectCli::GetRangeApp() noexcept
+NearObjectCli::GetServiceApp() noexcept
 {
-    return *m_rangeApp;
+    return *m_serviceApp;
 }
 
 CLI::App&
-NearObjectCli::GetRawApp() noexcept
+NearObjectCli::GetUwbRawApp() noexcept
 {
-    return *m_rawApp;
+    return *m_uwbRawApp;
 }
 
 CLI::App&
-NearObjectCli::GetRangeStartApp() noexcept
+NearObjectCli::GetUwbRangeApp() noexcept
 {
-    return *m_rangeStartApp;
+    return *m_uwbRangeApp;
 }
 
 CLI::App&
-NearObjectCli::GetRangeStopApp() noexcept
+NearObjectCli::GetServiceRangeApp() noexcept
 {
-    return *m_rangeStopApp;
+    return *m_serviceRangeApp;
+}
+
+CLI::App&
+NearObjectCli::GetUwbRangeStartApp() noexcept
+{
+    return *m_uwbRangeStartApp;
+}
+
+CLI::App&
+NearObjectCli::GetUwbRangeStopApp() noexcept
+{
+    return *m_uwbRangeStopApp;
+}
+
+CLI::App&
+NearObjectCli::GetServiceRangeStartApp() noexcept
+{
+    return *m_serviceRangeStartApp;
+}
+
+CLI::App&
+NearObjectCli::GetServiceRangeStopApp() noexcept
+{
+    return *m_serviceRangeStopApp;
 }
 
 std::shared_ptr<uwb::UwbDevice>
@@ -135,6 +159,7 @@ NearObjectCli::CreateParser() noexcept
 
     // sub-commands
     m_uwbApp = AddSubcommandUwb(app.get());
+    m_serviceApp = AddSubcommandService(app.get());
 
     return app;
 }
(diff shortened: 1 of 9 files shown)


corbin-phipps:
Current CLI output (with GetAppConfigParams code commented out of HandleDriverStartRanging):
image

tools/cli/include/nearobject/cli/NearObjectCliData.hxx
```diff
@@ -13,6 +13,15 @@
 
 namespace nearobject::cli
 {
+/**
+ * @brief Helper structure to hold parameters for a ranging session.
+ */
+struct UwbRangingParameters
+{
+    uint32_t sessionId;
+    std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter> appConfigParams;
```
abeltrano, line 22: nit: for structures that are purely data containers, we prefer to give the properties capitalized names. Also, please use complete words:

```suggestion
    uint32_t SessionId;
    std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter> ApplicationConfigurationParameters;
```

windows/devices/uwb/UwbCxAdapterDdiLrp.cxx
```diff
@@ -644,6 +644,15 @@ windows::devices::uwb::ddi::lrp::From(const UwbApplicationConfigurationParameter
             applicationConfigurationParameterWrapper = std::make_unique<UwbApplicationConfigurationParameterWrapper>(totalSize);
             UWB_APP_CONFIG_PARAM &applicationConfigurationParameter = applicationConfigurationParameterWrapper->value();
             applicationConfigurationParameter.paramValue[0] = value;
+        } else if constexpr (std::is_same_v<T, std::unordered_set<::uwb::UwbMacAddress>>) {
```
abeltrano, line 647: Thanks for adding this. Can you please add a test for this case in [`TestUwbCxAdapterDdiLrpConversion.cxx`](https://github.com/microsoft/nearobject-framework/blob/develop/tests/unit/windows/TestUwbCxAdapterDdiLrpConversion.cxx)?

tools/cli/NearObjectCli.cxx
```diff
@@ -237,63 +274,221 @@ NearObjectCli::AddSubcommandUwbMonitor(CLI::App* parent)
...
+        m_cliData->appConfigParamsData.deviceMacAddress = uwb::UwbMacAddress::FromString(m_cliData->deviceMacAddressString, macAddressType);
+        m_cliData->appConfigParamsData.destinationMacAddresses.emplace();
+        // TODO: Insert the correct Controlee mac address rather than simply inserting the entire destinationMacAddressString
+        for (auto i = 0; i < m_cliData->appConfigParamsData.numberOfControlees; i++) {
+            m_cliData->appConfigParamsData.destinationMacAddresses.value().insert(uwb::UwbMacAddress::FromString(m_cliData->destinationMacAddressString, macAddressType).value());
+        }
+
+        std::cout << "Selected parameters:" << std::endl;
+        // TODO: Display selected parameters
+        std::cout << "DeviceRole: " << static_cast<int>(m_cliData->appConfigParamsData.deviceRole.value()) << std::endl;
```
abeltrano, line 423: For each of these, why not display the enum name instead of the numerical integer value? eg. `magic_enum::enum_name(m_cliData->appConfigParamsData.deviceRole.value())`

tools/cli/NearObjectCli.cxx
```diff
@@ -237,63 +274,221 @@ NearObjectCli::AddSubcommandUwbMonitor(CLI::App* parent)
...
+    rangeStartApp->add_option("--DeviceMacAddress", m_cliData->deviceMacAddressString, deviceMacAddressDescription)->capture_default_str()->required();
+    rangeStartApp->add_option("--DestinationMacAddress", m_cliData->destinationMacAddressString, dstMacAddressDescription)->capture_default_str()->required();
+
+    detail::AddEnumOption(rangeStartApp, appConfigParamsData.deviceType, true);
+
+    rangeStartApp->parse_complete_callback([this, rangeStartApp] {
+        // TODO: Move validation logic into its own function
+
+        // Validate NumberOfControlees
+        if (m_cliData->appConfigParamsData.multiNodeMode == uwb::protocol::fira::MultiNodeMode::Unicast) {
```
abeltrano, line 402: Being that this namespace is used so much in this file, might as well add a `using namespace ::uwb::protocol::fira` at the top of this file and eliminate the prefixes used throughout.

tools/cli/NearObjectCli.cxx
```diff
@@ -237,63 +274,221 @@ NearObjectCli::AddSubcommandUwbMonitor(CLI::App* parent)
...
+            if (m_cliData->appConfigParamsData.numberOfControlees != 1) {
+                std::cerr << "Only 1 controlee expected in Unicast mode" << std::endl;
+            }
+        } else {
+            if (m_cliData->appConfigParamsData.numberOfControlees < 1 || m_cliData->appConfigParamsData.numberOfControlees > 8) {
```
abeltrano, line 407: Let's add a constant somewhere (probably `FiraDevice.hxx`) for max controlees so we can avoid magic numbers where possible, eg.

`constexpr auto MaxNumberOfControlees = 8`

Also, I might vaguely remember the spec says implementations may optionally support more than 8. If so, I wonder if we should not range check the inputs here and just pass the valeus onto the driver and let it do the checking?

abeltrano (changes requested):

tools/cli/NearObjectCli.cxx
corbin-phipps: Ah yeah I was thinking that that would be better but I forgot about magic_enum. Thanks!

tools/cli/NearObjectCli.cxx
corbin-phipps: Hmm I don't remember that, at least not in the UCI spec. I'd like to keep the check here for now, but I'll definitely add the constant

abeltrano: FYI, this is from CSML, table 7:
image

corbin-phipps: Oh I see, I can remove the check then

PR #173, opened 2023-03-27
Fix session callbacks issue in nocli

corbin-phipps:
### Type

- [x] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR fixes an issue where session callbacks for ranging data were not firing.

### Technical Details

- Changed UwbSession constructor to take a UwbDevice pointer instead of std::shared_ptr<UwbDeviceConnector>.
- Create new uwb device connector in UwbSession to register session event callbacks and call NotificationListenerStart in UwbSession.
- Save NearObjectCliSessionEventCallbacks in NearObjectCliHandler.
- Update outdated comment for NotificationListenerStart.

### Test Results

Ran nocli with TestClient driver and can successfully get multiple range data notifications without "missing callback" warnings.

### Reviewer Focus

None.

### Future Work

- Update UwbDevice to be created with shared references to avoid passing in a raw UwbDevice pointer

### Checklist

### DIFF
--- a/tools/cli/NearObjectCliHandler.cxx
+++ b/tools/cli/NearObjectCliHandler.cxx
@@ -12,14 +12,18 @@
 using namespace nearobject::cli;
 using namespace uwb::protocol::fira;
 
-NearObjectCliHandler::NearObjectCliHandler() :
-    m_sessionEventCallbacks(std::make_shared<NearObjectCliUwbSessionEventCallbacks>())
-{}
-
 void
 NearObjectCliHandler::SetParent(NearObjectCli* parent)
 {
     m_parent = parent;
+
+    auto controlFlowContext = (m_parent != nullptr) ? m_parent->GetControlFlowContext() : nullptr;
+    m_sessionEventCallbacks = std::make_shared<NearObjectCliUwbSessionEventCallbacks>([this, controlFlowContext = std::move(controlFlowContext)]() {
+        if (controlFlowContext != nullptr) {
+            controlFlowContext->OperationSignalComplete();
+        }
+        m_activeSession.reset();
+    });
 }
 
 std::shared_ptr<uwb::UwbDevice>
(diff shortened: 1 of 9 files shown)


tools/cli/NearObjectCliHandler.cxx
```diff
@@ -17,6 +16,14 @@ void
...
 {
     m_parent = parent;
+
+    auto controlFlowContext = (m_parent != nullptr) ? m_parent->GetControlFlowContext() : nullptr;
+    m_sessionEventCallbacks = std::make_shared<NearObjectCliUwbSessionEventCallbacks>([this, controlFlowContext = std::move(controlFlowContext)]() {
+        if (controlFlowContext != nullptr) {
+            controlFlowContext->OperationSignalComplete();
+        }
+        m_activeSession.reset();
+    });
```
abeltrano, line 26: Even though the callbacks make use of data obtained from the parent, the callbacks should still be created if `SetParent()` is not called. So, I think the assignment to `m_sessionEventCallbacks` needs to go somewhere else. I don't see why it can't be done in the c'tor; the lambda passed to the `NearObjectCliUwbSessionEventCallbacks` c'tor will capture local vars like `parent` and will resolve their values at the time the lambda is invoked.

However, what's here is good enough for what we need right now, so let's revisit this later (so go ahead and resolve this comment without taking any action).

abeltrano (approved):
Let's get this in after #165 is completed.

windows/devices/uwb/UwbSession.cxx
```diff
@@ -15,9 +15,8 @@
 using namespace windows::devices::uwb;
 using namespace ::uwb::protocol::fira;
 
-UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<::uwb::UwbSessionEventCallbacks> callbacks, std::shared_ptr<UwbDeviceConnector> uwbDeviceConnector, ::uwb::protocol::fira::DeviceType deviceType) :
-    ::uwb::UwbSession(sessionId, std::move(callbacks), deviceType),
-    m_uwbDeviceConnector(std::move(uwbDeviceConnector))
+UwbSession::UwbSession(uint32_t sessionId, std::weak_ptr<::uwb::UwbSessionEventCallbacks> callbacks, UwbDevice* uwbDevice, ::uwb::protocol::fira::DeviceType deviceType) :
```
forwardpointer, line 18: could you run the auto format on this file to make the cpp-linter test pass?

PR #174, opened 2023-03-27
UWB Simulator driver fixes

abeltrano:
### Type

- [X] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Ensure the simulator driver correctly implements `IOCTL_UWB_GET_APP_CONFIG_PARAMS`.

### Technical Details

* Add `UwbSession::Destroy` function to allow clients to destroy the session at will.
* Add ability to register callbacks to run when the `nocli` tool is finished running.
* Register stop callback to stop ranging and destroy the active ranging session upon `nocli` exit.
* Fix `DeviceIoControl` off-by-one loop exit conditions.
* Fix `DeviceIoControl` buffer size check to use correct value `ERROR_MORE_DATA` instead of `ERROR_INSUFFICIENT_BUFFER`.
* Fix simulator driver code to update existing application configuration parameters.
* Fix simulator driver code that partitions application configuration parameters per session state (logic was flipped).
* Fix simulator driver dispatch functions to complete requests with insufficient buffer size with `STATUS_BUFFER_OVERFLOW` instead of `STATUS_BUFFER_TOO_SMALL`. The latter is an error and so will not update the `lpBytesReturned` argument with the required number of bytes.
* Add output buffer size check for `UwbSimulatorDdiHandler::OnUwbGetApplicationConfigurationParameters`.
* Add documentation for `UwbDeviceConnector`.
* Add missing pieces for`UwbApplicationConfigurationParameterValue` `ToString()`

### Test Results

* All unit tests pass on both Linux and Windows.
* Ran `uwb range --SessionId 1234 start --DeviceRole 1 --MultiNodeMode 0 --NumberOfControlees 1 --DeviceMacAddress 12:34 --DestinationMacAddress 67:89 --DeviceType 1` with the following output:

```
Selected parameters:
DeviceRole: Initiator
MultiNodeMode: Unicast
NumberOfControlees: 1
DeviceMacAddress: 12:34
DestinationMacAddresses: 67:89
DeviceType: Controller
Initialize
configure session with id 1234
ConfigureImpl
IOCTL_UWB_NOTIFICATION attempt #0 with 0-byte buffer
IOCTL_UWB_SESSION_INIT succeeded
IOCTL_UWB_SET_APP_CONFIG_PARAMS succeeded
get application configuration parameters
IOCTL_UWB_GET_APP_CONFIG_PARAMS attempt #0 with 0-byte buffer
IOCTL_UWB_GET_APP_CONFIG_PARAMS attempt #1 with 92-byte buffer
IOCTL_UWB_GET_APP_CONFIG_PARAMS succeeded
Session Application Configuration Parameters:
 > DeviceType: Controller
 > MultiNodeMode: Unicast
 > NumberOfControlees: 1
 > DeviceMacAddress: 12:34
 > DestinationMacAddresses: 67:89
 > DeviceRole: Initiator
start ranging
IOCTL_UWB_START_RANGING_SESSION succeeded
```

### Reviewer Focus

None

### Future Work

* Allow overriding close-session-on-exit using a `nocli` flag.

### Checklist

### DIFF
--- a/tools/cli/include/nearobject/cli/NearObjectCliHandler.hxx
+++ b/tools/cli/include/nearobject/cli/NearObjectCliHandler.hxx
@@ -7,9 +7,9 @@
 
 #include <nearobject/cli/NearObjectCliControlFlowContext.hxx>
 #include <nearobject/cli/NearObjectCliData.hxx>
-#include <uwb/UwbSessionEventCallbacks.hxx>
 #include <uwb/UwbDevice.hxx>
 #include <uwb/UwbSession.hxx>
+#include <uwb/UwbSessionEventCallbacks.hxx>
 #include <uwb/protocols/fira/UwbSessionData.hxx>
 
 namespace nearobject::cli
--- a/tools/cli/include/nearobject/cli/NearObjectCliHandler.hxx
+++ b/tools/cli/include/nearobject/cli/NearObjectCliHandler.hxx
@@ -7,6 +7,7 @@
 
 #include <nearobject/cli/NearObjectCliControlFlowContext.hxx>
 #include <nearobject/cli/NearObjectCliData.hxx>
+#include <uwb/UwbSessionEventCallbacks.hxx>
 #include <uwb/UwbDevice.hxx>
 #include <uwb/UwbSession.hxx>
 #include <uwb/protocols/fira/UwbSessionData.hxx>
@@ -99,6 +100,7 @@ struct NearObjectCliHandler
 private:
     NearObjectCli* m_parent;
     std::shared_ptr<::uwb::UwbSession> m_activeSession;
+    std::shared_ptr<::uwb::UwbSessionEventCallbacks> m_sessionEventCallbacks;
 };
 
 } // namespace nearobject::cli
--- a/windows/devices/uwb/UwbDeviceConnector.cxx
+++ b/windows/devices/uwb/UwbDeviceConnector.cxx
@@ -103,7 +103,7 @@ UwbDeviceConnector::GetDeviceInformation()
     // the case, the first attempt will succeed. Otherwise, the buffer is grown
     // to account for the vendor specific information, and the IOCTL attempted a
     // second time.
-    for (const auto i : std::ranges::iota_view{ 1, 2 }) {
+    for (const auto i : std::ranges::iota_view{ 0, 2 }) {
         deviceInformationBuffer.resize(bytesRequired);
         PLOG_DEBUG << "IOCTL_UWB_GET_DEVICE_INFO attempt #" << i << " with " << std::size(deviceInformationBuffer) << "-byte buffer";
         BOOL ioResult = DeviceIoControl(handleDriver.get(), IOCTL_UWB_GET_DEVICE_INFO, nullptr, 0, std::data(deviceInformationBuffer), std::size(deviceInformationBuffer), &bytesRequired, nullptr);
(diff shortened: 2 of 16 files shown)


tools/cli/include/nearobject/cli/NearObjectCliHandler.hxx
```diff
@@ -99,8 +100,9 @@ struct NearObjectCliHandler
 
 private:
     NearObjectCli* m_parent;
+    std::shared_ptr<::uwb::UwbDevice> m_activeDevice;
     std::shared_ptr<::uwb::UwbSession> m_activeSession;
-    std::shared_ptr<NearObjectCliUwbSessionEventCallbacks> m_sessionEventCallbacks;
+    std::shared_ptr<::uwb::UwbSessionEventCallbacks> m_sessionEventCallbacks;
```
corbin-phipps, line 105: Why don't we want to use NearObjectCliUwbSessionEventCallbacks here?

forwardpointer (dismissed):

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -103,14 +103,14 @@ UwbConnector::GetDeviceInformation()
...
-    for (const auto i : std::ranges::iota_view{ 1, 2 }) {
+    for (const auto i : std::ranges::iota_view{ 0, 2 }) {
         deviceInformationBuffer.resize(bytesRequired);
         PLOG_DEBUG << "IOCTL_UWB_GET_DEVICE_INFO attempt #" << i << " with " << std::size(deviceInformationBuffer) << "-byte buffer";
         BOOL ioResult = DeviceIoControl(handleDriver.get(), IOCTL_UWB_GET_DEVICE_INFO, nullptr, 0, std::data(deviceInformationBuffer), std::size(deviceInformationBuffer), &bytesRequired, nullptr);
         if (!LOG_IF_WIN32_BOOL_FALSE(ioResult)) {
             DWORD lastError = GetLastError();
             // Treat all errors other than insufficient buffer size as fatal.
-            if (lastError != ERROR_INSUFFICIENT_BUFFER) {
+            if (lastError != ERROR_MORE_DATA) {
```
corbin-phipps, line 113: Could we perhaps leave these changes for another PR? This would require immediate changes to the CX, which I'd like to avoid until we can have a discussion with Davut as well

corbin-phipps (approved):

tools/cli/include/nearobject/cli/NearObjectCliHandler.hxx
forwardpointer: I'm guessing its because NearObjectCliUwbSessionEventCallbacks is the derived class and doesn't need to be part of the public interface

abeltrano: `NearObjectCliUwbSessionEventCallbacks` is just an implementation of the interface that's actually required, `::uwb::UwbSessionEventCallbacks`. How it implements it (and the additional stuff it exposes, if any) is just an implementation detail that `NearObjectCliHandler.cxx` should know the minimum required about. Typically, storing an instance to the least-derived class in a hierarchy is safest since it reduces, and in many cases, eliminates the possible interactions with an instance, which means it's easier to understand and the domain of test outputs is minimized. In this case, `NearObjectClieHandler` only needs to know how to construct the instance with the specific bits it needs (the session ended callback), so there's no need to store a more specific derived instance.

windows/devices/uwb/UwbDeviceConnector.cxx
abeltrano: Yep, I started a discussion offline. We'll want to close on this quickly, though, since we rely on the simulator for testsing, and this is going to become increasingly important once we move up the stack toward the true service portions.

abeltrano: Tagging @davutkarabay.

abeltrano: This one is synchronous, so I've left it. However, I reverted the one for `IOCTL_UWB_NOTIFICATION` which is overlapped/async, so will resolve this comment.

PR #181, opened 2023-03-29
Add ability to obtain session information directly from driver layer

abeltrano:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* Allow creating a `UwbSession` from a session identifier alone.
* Allow service layer to query the driver layer for active session information.

### Technical Details
 
 * Expose `UwbSession::GetSession()` function which allows clients with a-priori information to obtain a `UwbSession` reference.
 * Add `UwbSession::ResolveSessionImpl()` function which allows higher-layer code to resolve pre-existing sessions from the driver layer.
* Add ability to retrieve and change the `UwbSession` event callbacks.
* Add helper to resolve session event callbacks form derived classes.
* Update session event callbacks checks to avoid implicit conversion to `bool`.
 
### Test Results

Performed some ad-hoc scenarios exercising full ranging using the simulator driver and ensured there were no regressions.

### Reviewer Focus

None

### Future Work

* The Windows derived `ResolveSessionImpl()` must be implemented. This requires allowing access to `IUwbSessionDdi` through the device connector, which is not currently possible without changing the containing type.
* The `GetSession()` function will be used from the raw commands involving sessions.

### Checklist

### DIFF
--- a/windows/devices/uwb/UwbDevice.cxx
+++ b/windows/devices/uwb/UwbDevice.cxx
@@ -47,6 +47,14 @@ UwbDevice::CreateSessionImpl(uint32_t sessionId, std::weak_ptr<::uwb::UwbSession
 std::shared_ptr<::uwb::UwbSession>
 UwbDevice::ResolveSessionImpl([[maybe_unused]] uint32_t sessionId)
 {
+    // TODO: implement this
+    // One problem is that m_uwbDeviceConnector is of type
+    // IUwbDeviceDdiConnector, which prevents use of the session-based connector
+    // functionality. That is needed here to obtain information about the
+    // session in question. For example, we need to call
+    // IUwbSessionDdi::[SessionGetState, GetApplicationConfigurationParameters,
+    // SessionGetRangingCount]. So, we may need to change the type of
+    // m_uwbDeviceConnector to UwbConnector instead.
     return nullptr;
 }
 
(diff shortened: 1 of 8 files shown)


windows/devices/uwb/UwbDevice.cxx
```diff
@@ -38,12 +38,26 @@ UwbDevice::DeviceName() const noexcept
...
+{
+    // TODO: implement this
+    // One problem is that m_uwbDeviceConnector is of type
+    // IUwbDeviceDdiConnector, which prevents use of the session-based connector
+    // functionality. That is needed here to obtain information about the
+    // session in question. For example, we need to call
+    // IUwbSessionDdi::[SessionGetState, GetApplicationConfigurationParameters,
+    // SessionGetRangingCount]. So, we may need to change the type of
+    // m_uwbDeviceConnector to UwbConnector instead.
+    return nullptr;
```
forwardpointer, line 58: I think maybe you can do a dynamic cast here so that the UwbDevice.hxx interface still shows the m_uwbDeviceConnector as IUwbDeviceDdiConnector, or maybe it doesn't really matter since m_uwbDeviceConnector is private anyways

forwardpointer (approved):

windows/devices/uwb/UwbDevice.cxx
abeltrano: We could. I just thought of another idea though, where we can store once instance of each interface. I think that would be cleaner.

PR #192, opened 2023-03-31
Improve app configuration parameter display and vector building code in nocli

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [x] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR improves the code that displays the selected application configuration parameters and constructs a vector of these parameters.

### Technical Details

- Add ParameterTypesVariant and GetValueMap() to UwbApplicationConfigurationParameterData struct in NearObjectCliData.hxx 
- Move parameter displaying/parameter vector building code into its own function
- Improve display/vector building code to use new map instead of manually handling each parameter individually

### Test Results

nocli still works and displays the parameters as expected.

### Reviewer Focus

None.

### Future Work

None.

### Checklist

### DIFF
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -231,6 +231,42 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory
     }
 }
 
+/**
+ * @brief Displays and returns application configuration parameters.
+ *
+ * @param applicationConfigurationParameterData The possible application configuration parameters that may be set.
+ * @return std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter>.
+ */
+std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter>
+ProcessApplicationConfigurationParameters(UwbApplicationConfigurationParameterData applicationConfigurationParameterData)
+{
+    std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter> applicationConfigurationParameters;
+
+    std::cout << "Selected parameters:" << std::endl;
+    for (const auto& [applicationConfigurationParameter, applicationConfigurationParameterValue] : applicationConfigurationParameterData.GetValueMap()) {
+        std::visit([&applicationConfigurationParameter, &applicationConfigurationParameters](auto&& arg)
+        {
+            applicationConfigurationParameters.push_back({ applicationConfigurationParameter, arg });
+
+            using ParameterValueT = std::decay_t<decltype(arg)>;
+            if constexpr (std::is_enum_v<ParameterValueT>) {
+                std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << magic_enum::enum_name(arg) << std::endl;
+            } else if constexpr (std::is_same_v<ParameterValueT, uint8_t>) {
+                std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << static_cast<int>(arg) << std::endl;
+            } else if constexpr (std::is_same_v<ParameterValueT, ::uwb::UwbMacAddress>) {
+                std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << ToString(arg) << std::endl;
+            } else if constexpr (std::is_same_v<ParameterValueT, std::unordered_set<::uwb::UwbMacAddress>>) {
+                for (const auto& address : arg) {
+                    std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << ToString(arg) << std::endl;
+                }
+            } 
+        },
+            applicationConfigurationParameterValue);
+    }
+
+    return applicationConfigurationParameters;
+}
+
 } // namespace detail
 
 CLI::App*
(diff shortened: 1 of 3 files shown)


tools/cli/include/nearobject/cli/NearObjectCliData.hxx
```diff
@@ -93,6 +95,46 @@ struct UwbApplicationConfigurationParameterData
...
+            valuesMap[UwbApplicationConfigurationParameterType::DeviceMacAddress] = deviceMacAddress.value();
+        }
+        if (destinationMacAddresses.has_value()) {
+            valuesMap[UwbApplicationConfigurationParameterType::DestinationMacAddresses] = destinationMacAddresses.value();
+        }
+        if (deviceType.has_value()) {
+            valuesMap[UwbApplicationConfigurationParameterType::DeviceType] = deviceType.value();
+        }
+        return valuesMap;
+    }
```
abeltrano, line 137: Can you make this a declaration and move the definition into `NearObjectCliData.cxx` please?

In general, please avoid putting code in headers. There are a whole bunch of good reasons for this: separating the implementation from the interface, ease of debugging, reducing build times, etc. The only cases I can think of where it's reasonable to have code in a header is for templates, `constexpr` or `consteval` expressions, and `default`, compiler-generated member functions; all of those are required by the standard to be in the header.

tools/cli/NearObjectCli.cxx
```diff
@@ -231,6 +231,42 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory
...
+    for (const auto& [applicationConfigurationParameter, applicationConfigurationParameterValue] : applicationConfigurationParameterData.GetValueMap()) {
+        std::visit([&applicationConfigurationParameter, &applicationConfigurationParameters](auto&& arg)
+        {
+            applicationConfigurationParameters.push_back({ applicationConfigurationParameter, arg });
+
+            using ParameterValueT = std::decay_t<decltype(arg)>;
+            if constexpr (std::is_enum_v<ParameterValueT>) {
+                std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << magic_enum::enum_name(arg) << std::endl;
+            } else if constexpr (std::is_same_v<ParameterValueT, uint8_t>) {
+                std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << static_cast<int>(arg) << std::endl;
```
abeltrano, line 255: Even though it's not an issue in this instance, we should use unary integer promotion here instead as good practice:

```suggestion
                std::cout << magic_enum::enum_name(applicationConfigurationParameter) << "::" << +arg << std::endl;
```

The presence of the cast might create a tendency to do that elsewhere in cases where a narrowing conversion would occur, and truncate data.

tools/cli/NearObjectCli.cxx
```diff
@@ -231,6 +231,42 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory
...
+ * @param applicationConfigurationParameterData The possible application configuration parameters that may be set.
+ * @return std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter>.
+ */
+std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter>
+ProcessApplicationConfigurationParameters(UwbApplicationConfigurationParameterData applicationConfigurationParameterData)
+{
+    std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter> applicationConfigurationParameters;
+
+    std::cout << "Selected parameters:" << std::endl;
+    for (const auto& [applicationConfigurationParameter, applicationConfigurationParameterValue] : applicationConfigurationParameterData.GetValueMap()) {
```
abeltrano, line 246: Could you put the value map into its own variable? The compiler is going to do this anyway, but having an explicit variable makes it far easier to inspect the values when running under a debugger:

```suggestion
    const auto applicationConfigurationParameterValues = applicationConfigurationParameterData.GetValueMap();
    for (const auto& [applicationConfigurationParameter, applicationConfigurationParameterValue] : applicationConfigurationParameterValues) {
```

abeltrano (changes requested):
Would you mind adding a screenshot or text-grab of the updated output?

tools/cli/include/nearobject/cli/NearObjectCliData.hxx
```diff
@@ -11,6 +11,8 @@
 #include <uwb/protocols/fira/UwbConfiguration.hxx>
 #include <uwb/protocols/fira/UwbSessionData.hxx>
 
+using namespace uwb::protocol::fira;
```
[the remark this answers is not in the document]
corbin-phipps: Oh this is good to know, thanks! I'll update the code

tools/cli/include/nearobject/cli/NearObjectCliData.hxx
corbin-phipps: Sure, I totally overlooked that this was in the header. Will move

corbin-phipps:
Updated output:
image

abeltrano (approved):

PR #193, opened 2023-03-31
Fix control flow of notification processing loop

abeltrano:
### Type

- [X] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Ensure that only valid notification data is processed from the driver.

The existing code was updated to use a `switch` statement, with control statements like `break` and `continue` unchanged. Those statement were intended to apply to the outer `for` loop, and so when used within the `switch`, are constrained to that scope. This caused the loop not to exit when the I/O operation is aborted, which then caused the code to process an empty buffer, which created corrupt data and eventually, a crash.

### Technical Details

* Replace `switch` with direct `if` statements.
* Improve logging.

### Test Results

* Ran `nocli.exe` for a complete ranging session, then a separate instance with the `sessiondeinit` command and observed the second instance complete successfully.

### Reviewer Focus

None.

### Future Work

The I/O code using `GetOverlappedResult` needs to eventually be replaced with threadpools.

### Checklist

### DIFF
--- a/windows/devices/uwb/UwbDeviceConnector.cxx
+++ b/windows/devices/uwb/UwbDeviceConnector.cxx
@@ -506,7 +506,7 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
                             continue;
                         } else if (lastError == ERROR_OPERATION_ABORTED) {
                             LOG_WARNING << "IOCTL_UWB_NOTIFICATION aborted";
-                            break; // for({0,2})                        
+                            break; // for({0,2})
                         } else {
                             PLOG_ERROR << "error when sending IOCTL_UWB_NOTIFICATION, hr=" << std::showbase << std::hex << hr;
                             break; // for({0,2})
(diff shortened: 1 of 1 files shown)


windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -490,33 +490,35 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
...
                         lastError = GetLastError();
-                        HRESULT hr = HRESULT_FROM_WIN32(lastError);
-                        switch (lastError) {
-                        case ERROR_INSUFFICIENT_BUFFER: // occurs for overlapped i/o
-                        case ERROR_MORE_DATA:           // occurs for synchronous i/o
-                            LOG_DEBUG << "IOCTL_UWB_NOTIFICATION completed, hr=" << std::showbase << std::hex << hr;
-                            // driver indicated buffer was too small and put required size in 'bytesRequired'. Retry with new size.
+                        hr = HRESULT_FROM_WIN32(lastError);
+                        if (lastError == ERROR_INSUFFICIENT_BUFFER || lastError == ERROR_MORE_DATA) {
+                            // Driver indicated buffer was too small and put required size in 'bytesRequired'. Retry with new size.
```
corbin-phipps, line 504: This flow may need to be updated now that we have new information from Wei regarding the DeviceIoControl. The required size should not be put in bytesRequired anymore - it should be set in the "size" field of the structs. This change has been made in the CX but should also be made in the Simulator driver I'm assuming

corbin-phipps (changes requested):

windows/devices/uwb/UwbDeviceConnector.cxx
abeltrano: Yes, it'll have to be updated, but that will be done in another PR.

PR #194, opened 2023-03-31
Add support for handling multiple UwbNotificationData events in simulator driver

abeltrano:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Allow > 1 simulator clients to listen for notifications.

The notification processing design was overhauled significantly. Previously, there was a global i/o event queue attached to the `UwbSimulatorDevice` instance. This made it difficult to service multiple open-file handles without additional context, and open-file handles are already managed by the WDF framework for which we have a dedicated context object, `UwbSimulatorDeviceFile`.

The design goals are to:
1. Allow each open-file handle to wait on a single `IOCTL_UWB_NOTIFICATION` event.
2. Allow each open-file handle to queue uwb notification data when not actively awaiting on the handle.

The design was updated to have each open-file handle have and manage its own i/o event queue, becoming an i/o event sink. This allows the file handle context to queue notification data when the client is not actively waiting on the `IOCTL_UWB_NOTIFICATION` ioctl, for example, when the client is processing the last event notification data.

#### UwbDeviceIoEventQueue
The i/o event queue, `UwbDeviceIoEventQueue`, is composed of 2 queues: one to queue `UwbNotificationData` entries, called the data queue, and one to queue the single pended `WDFREQUEST` from `DeviceIoControl`, called the request queue. Each queue has a corresponding condition variable associated with it that is used to wait and signal the resource state (request available, notification available). Both condition variables share the same mutex to synchronize access to the shared data and condition variable state. The i/o event queue uses a dedicated thread to wait for the condition variables to be signaled. The high level algorithm is that the processing thread first waits for a request to be pended, then waits for notification data to become available. Once available, it completes the pended `WDFREQUEST`, and repeats until exit is requested.

Since it is expected the client uses the open-file handle for dedicated event processing, the data queue is artificially limited in size to 16 `UwbNotificationData` entries. This may be adjusted in future based on empirical data showing whether that queue size is properly sized.

The `UwbSimulatorDevice` newly exposes functionality to accept uwb notification data, which is then distributed to each open-file handle queue, becoming an i/o event source. This required updating `UwbSimulatorDevice` to track open-file handle contexts (`UwbSimulatorDeviceFile`).

### Technical Details

* Remove i/o event queue from `UwbSimulatorDevice`.
* Track open file handle pointers in `UwbSimulatorDevice`.
* Pend all notification requests unconditionally to simplify processing logic.

### Test Results

* Ran end-to-end ranging scenario and validated state changes and notification data was delivered to `nocli.exe` client.

### Reviewer Focus

* Concurrency here is complicated so it deserves a very close look. In particular, consider corner cases where a deadlock could occur with the condition variables used to wait/signal notification event data and pended request availability in `UwbSimulatorIoEventQueue`.

### Future Work

* Eliminate copies of `UwbNotificationData` being passed around to each open-file i/o event queue.
* Switch to using smart pointers for most objects to manage lifetimes more safely.

### Checklist

### DIFF
--- a/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
+++ b/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
@@ -200,7 +200,7 @@ UwbSimulatorDevice::OnFileClose(WDFFILEOBJECT file)
         return (uwbSimulatorFile == uwbSimulatorFileClosed);
     });
 
-    DbgPrint("%p %s file object %p", m_wdfDevice, (removed ? "removed" : "failed to remove"), file);
+    DbgPrint("%p %s file object %p\n", m_wdfDevice, (removed ? "removed" : "failed to remove"), file);
 }
 
 NTSTATUS
--- a/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
+++ b/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
@@ -172,7 +172,7 @@ UwbSimulatorDevice::OnFileCreate(WDFDEVICE device, WDFREQUEST request, WDFFILEOB
         TraceLoggingPointer(request, "Request"),
         TraceLoggingPointer(file, "File"));
 
-    auto uwbSimulatorFileBuffer = GetUwbSimulatorFile(file);
+    auto uwbSimulatorFileBuffer = GetUwbSimulatorDeviceFile(file);
     auto uwbSimulatorFile = new (uwbSimulatorFileBuffer) UwbSimulatorDeviceFile(file, this);
     auto uwbSimulatorFileStatus = uwbSimulatorFile->Initialize();
     if (uwbSimulatorFileStatus == STATUS_SUCCESS) {
@@ -198,7 +198,7 @@ UwbSimulatorDevice::OnFileClose(WDFFILEOBJECT file)
         TraceLoggingPointer(file, "File"));
 
     std::unique_lock deviceFilesLockExclusive{ m_deviceFilesGate };
-    const auto uwbSimulatorFileClosed = GetUwbSimulatorFile(file);
+    const auto uwbSimulatorFileClosed = GetUwbSimulatorDeviceFile(file);
     const auto removed = std::erase_if(m_deviceFiles, [&](const auto &uwbSimulatorFile) {
         return (uwbSimulatorFile == uwbSimulatorFileClosed);
     });
--- a/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
+++ b/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
@@ -183,6 +183,7 @@ UwbSimulatorDevice::OnFileCreate(WDFDEVICE device, WDFREQUEST request, WDFFILEOB
         DbgPrint("%p added file object %p\n", m_wdfDevice, file);
     } else {
         uwbSimulatorFile->~UwbSimulatorDeviceFile();
+        DbgPrint("%p failed to intialize file object context with status 0x%08x\n", status);
     }
 
     WdfRequestComplete(request, uwbSimulatorFileStatus);
--- a/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
+++ b/windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
@@ -111,36 +111,6 @@ UwbSimulatorDevice::OnWdfFileClose(WDFFILEOBJECT file)
 NTSTATUS
 UwbSimulatorDevice::Initialize()
 {
-    std::shared_ptr<UwbSimulatorIoEventQueue> ioEventQueue;
-    {
-        // Create a manual dispatch queue for event handling.
-        WDF_IO_QUEUE_CONFIG queueConfig;
-        WDF_IO_QUEUE_CONFIG_INIT(&queueConfig, WdfIoQueueDispatchManual);
-        queueConfig.PowerManaged = WdfFalse;
-
-        WDFQUEUE wdfQueue;
-        NTSTATUS status = WdfIoQueueCreate(m_wdfDevice, &queueConfig, WDF_NO_OBJECT_ATTRIBUTES, &wdfQueue);
-        if (!NT_SUCCESS(status)) {
-            TraceLoggingWrite(
-                UwbSimulatorTraceloggingProvider,
-                "WdfIoQueueCreate failed",
-                TraceLoggingLevel(TRACE_LEVEL_FATAL),
-                TraceLoggingNTStatus(status));
-            return status;
-        }
-
-        // Construct a new UwbSimulatorIoEventQueue instance using the WDF pre-allocated buffer.
-        ioEventQueue = std::make_shared<UwbSimulatorIoEventQueue>(wdfQueue);
-        if (ioEventQueue == nullptr) {
-            return STATUS_INSUFFICIENT_RESOURCES;
-        }
-
-        status = ioEventQueue->Initialize();
-        if (!NT_SUCCESS(status)) {
-            return status;
-        }
-    }
-
     // Create a default queue for all requests that are not explicitly configured to go elsewhere.
     WDF_IO_QUEUE_CONFIG queueConfig;
     WDF_IO_QUEUE_CONFIG_INIT_DEFAULT_QUEUE(&queueConfig, WdfIoQueueDispatchParallel);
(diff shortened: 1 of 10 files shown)


windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
```diff
@@ -209,11 +172,15 @@ UwbSimulatorDevice::OnFileCreate(WDFDEVICE device, WDFREQUEST request, WDFFILEOB
...
+    auto uwbSimulatorFileBuffer = GetUwbSimulatorDeviceFile(file);
     auto uwbSimulatorFile = new (uwbSimulatorFileBuffer) UwbSimulatorDeviceFile(file, this);
     auto uwbSimulatorFileStatus = uwbSimulatorFile->Initialize();
     if (uwbSimulatorFileStatus == STATUS_SUCCESS) {
         uwbSimulatorFile->RegisterHandler(m_ddiHandler);
+
+        std::unique_lock deviceFilesLockExclusive{ m_deviceFilesGate };
+        m_deviceFiles.push_back(uwbSimulatorFile);
+        DbgPrint("%p added file object %p\n", m_wdfDevice, file);
     } else {
```
corbin-phipps, line 184: Might be useful to add a log line in the "else" block

windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
```diff
@@ -209,11 +172,15 @@ UwbSimulatorDevice::OnFileCreate(WDFDEVICE device, WDFREQUEST request, WDFFILEOB
...
         uwbSimulatorFile->~UwbSimulatorDeviceFile();
```
corbin-phipps, line 185: I don't love this, but it seems like it might be necessary due to the placement new above? Is that necessary because of WDF reasons?

corbin-phipps (commented):
Looks good but I want to take another pass later. Lots of concurrency to look at!

windows/drivers/uwb/simulator/UwbSimulatorIoEventQueue.hxx
```diff
@@ -58,31 +61,34 @@ public:
...
-     * @brief Handle a WDF request for notification data.
-     *
-     * @param request The WDF request handle.
-     * @param notificationData The output argument to hold notification request data, if available.
-     * @param outputBufferSize The size of the DDI output buffer that will eventually be populated.
-     * @return NTSTATUS
-     */
     NTSTATUS
-    HandleNotificationRequest(WDFREQUEST request, std::optional<::uwb::protocol::fira::UwbNotificationData> &notificationData, std::size_t &outputBufferSize);
+    EnqueueRequest(WDFREQUEST request);
```
corbin-phipps, line 65: Missing function comment

corbin-phipps (changes requested):

windows/drivers/uwb/simulator/UwbSimulatorDeviceFile.hxx
```diff
@@ -86,6 +90,9 @@ private:
     WDFFILEOBJECT m_wdfFile;
     UwbSimulatorDevice *m_uwbSimulatorDevice{ nullptr };
     std::vector<std::shared_ptr<windows::devices::uwb::simulator::IUwbSimulatorDdiHandler>> m_ddiHandlers{};
+    std::shared_ptr<UwbSimulatorIoEventQueue> m_ioEventQueue;
+
+    static constexpr std::size_t MaximumQueueSizeDefault = 16;
```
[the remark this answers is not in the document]
corbin-phipps: Never mind, I see you answered my question in the PR description

corbin-phipps:
Are there unit tests to cover some of the complicated concurrency edge cases? Or are we just validating with simulated ranging data + nocli for now?

windows/drivers/uwb/simulator/UwbSimulatorDeviceFile.hxx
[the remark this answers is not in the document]
abeltrano: I should document it here as well for future refernece.

windows/drivers/uwb/simulator/UwbSimulatorIoEventQueue.cxx
```diff
@@ -39,83 +49,108 @@ UwbSimulatorIoEventQueue::Uninitialize()
...
-    auto notificationDataDdi = UwbCxDdi::From(notificationData);
-    auto notificationDataDdiBuffer = std::data(notificationDataDdi);
-    auto notificationDataDdiBufferSize = std::size(notificationDataDdi);
-
-    NTSTATUS status = WdfWaitLockAcquire(m_wdfQueueLock, nullptr);
-    if (!NT_SUCCESS(status)) {
-        return status;
+    // Push the notification data into the queue.
+    {
+        std::unique_lock notificationLockExclusive{ m_notificationGate };
```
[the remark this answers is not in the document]
abeltrano: `std::lock_guard` could be used here, yes. In that case, `std::scoped_lock` is almost always the right choice instead of `std::lock_guard` since it's essentially its successor. I originally used `std::unique_lock` here because the initial implementation used a standard `std::condition_variable` which requires a `std::unique_lock` and I wanted to use consistent RAII locking objects. However, we're now using std::condition_variable_any` which takes any lockable type, not just `std::unique_lock`, so, that should take the `std::scoped_lock` as well.

windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
```diff
@@ -318,3 +290,16 @@ UwbSimulatorDevice::GetSessionCount()
     std::shared_lock sessionsReadLock{ m_sessionsGate };
     return std::size(m_sessions);
 }
+
+void
+UwbSimulatorDevice::PushUwbNotification(UwbNotificationData uwbNotificationData)
+{
+    // TODO: log
+    // TODO: lock
```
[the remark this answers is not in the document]
abeltrano: I added a `std::mutex` for this since the first iteration and removed the TODO. 

But to your point: the locking is necessary here, irrespective of what happens in `UwbSimulatorIoEventQueue::PushNotification()` because we're reading `m_deviceFiles` which will be modified any time a handle to the driver is opened, so access to it needs to be synchronized.

However, looking at this again, we really shouldn't continue to hold the lock while executing `PushNotification` for each device file, since we'll already have obtained a `shared_ptr` at that point. I'll make an improvement here.

abeltrano:
> Are there unit tests to cover some of the complicated concurrency edge cases? Or are we just validating with simulated ranging data + nocli for now?

Per our offline conversation, we more or less decided not to develop separate tests for the simulator driver since its purpose is to help with testing the service code. Eventually, the plan is to run the tests for UwbCx against it.

windows/drivers/uwb/simulator/UwbSimulatorIoEventQueue.hxx
abeltrano: Will update, thanks!

windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
abeltrano: Are you referring to manually invoking the destructor? If so, you're correct that it has to be invoked manually because it was constructed with a contiguous buffer from the WDF framework object, which doesn't have strong type-safety and so won't invoke any kind of destructor on it when it goes out of scope.

windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
abeltrano: Good idea.

windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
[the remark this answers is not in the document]
abeltrano: Sorry, my comment is unintelligible. I was actually referring to changes I haven't sent yet which change the pointer type from raw to weak_ptr. I'll make this change on that next PR.

windows/drivers/uwb/simulator/UwbSimulatorDevice.cxx
corbin-phipps: Yes, that's what I was referring to, and your explanation reinforces my understanding. Thanks!

corbin-phipps (approved):

PR #199, opened 2023-04-03
Autostart the UwbConnector when the first event callback is registered

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [X] Infrastructure

### Goals

Autostart the UwbConnector when the first event callback is registered

### Technical Details

* There is a shared_mutex for reading/writing the callbacks weak_ptrs
* There is also a mutex and condition_variable for detecting the presence of callbacks

### Test Results

Verified that the basic ranging with random data scenario works with the simulator driver 

### Future Work

This design might need to be changed once we implement deregistering callbacks. 

### Checklist


forwardpointer:
I verified that the basic ranging scenario works with the simulator driver, but it's important to note that on the develop branch when I ran the same scenario the uwbsim.exe is able to exit by itself but not with this PR. This may be a bug

forwardpointer:
> I verified that the basic ranging scenario works with the simulator driver, but it's important to note that on the develop branch when I ran the same scenario the uwbsim.exe is able to exit by itself but not with this PR. This may be a bug

Ok I verified that this is caused by an unused UwbConnector member in the UwbSimulator, so I removed it and uwbsim.exe exits properly now

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -731,19 +739,36 @@ UwbConnector::NotificationListenerStop()
 ::uwb::RegisteredCallbackToken*
 UwbConnector::RegisterDeviceEventCallbacks(std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> callbacks)
 {
-    m_deviceEventCallbacks = callbacks;
+    {
+        std::lock_guard presenceLock{ m_callbacksPresentConditionVariableGate };
```
corbin-phipps, line 743: Consider using `std::scoped_lock` instead of `std::lock_guard`, as `std::scoped_lock` is the newer and superior locking mechanism

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -192,8 +195,23 @@ private:
...
+     * @brief Internal function to check if there are callbacks present
+     *
+     * @return true
+     * @return false
+     */
+    bool
+    CallbacksPresent();
+
 private:
+    mutable std::shared_mutex m_eventCallbacksGate;
```
corbin-phipps, line 208: Why is the `mutable` keyword needed?

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -192,8 +195,23 @@ private:
...
+
+    mutable std::mutex m_callbacksPresentConditionVariableGate;
+    mutable std::condition_variable m_callbacksPresentConditionVariable;
+
     std::unordered_map<uint32_t, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks>> m_sessionEventCallbacks;
+
```
corbin-phipps, line 214: nit: Looks like this newline was added, not sure if intentional

corbin-phipps (changes requested):

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
abeltrano: This is typically done such that non-const locking functions can be used in `const` functions. It's technically cheating since the lock state does change, but the idea of `const` is more meant to reflect actual object state and not pseudo-state used to control concurrency.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -29,7 +29,9 @@ class RegisteredCallbackToken
 
 UwbConnector::UwbConnector(std::string deviceName) :
     m_deviceName(std::move(deviceName))
-{}
+{
+    NotificationListenerStart();
```
abeltrano, line 33: Why is this needed here? It seemed like the callback registration code would take care of calling this when necessary.

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -192,8 +195,23 @@ private:
...
+     * @return true
+     * @return false
+     */
+    bool
+    CallbacksPresent();
+
 private:
+    mutable std::shared_mutex m_eventCallbacksGate;
+
+    mutable std::mutex m_callbacksPresentConditionVariableGate;
```
abeltrano, line 210: 2 separate mutexes aren't necessary since they're both protecting the same state. You can use the `std::shared_mutex` here as long as you change `m_callbacksPresentConditionVariable` to be a `std::condition_variable_any` and ensure that the lock you pass to `wait()` is a `std::unique_lock` of the `m_eventCallbacksGate`. When `wait()` is called, the mutex is atomically _released_, making it available to synchronize access to `m_sessionEventCallbacks` and `m_deviceEventCallbacks` in whatever way you want (ie. either with `std::shared_lock`, `std::unique_lock`, `std::scoped_lock`, etc.)

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
abeltrano: I personally like grouping the concurrency related members together.

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -150,6 +138,21 @@ public:
     SetApplicationConfigurationParameters(uint32_t sessionId, std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> applicationConfigurationParameters) override;
 
 private:
+    /**
+     * @brief Start listening for notifications.
+     *
+     * @return true If listening for notifications started successfully.
+     * @return false If listening for notifications could not be started.
+     */
+    bool
```
abeltrano, line 147: It made sense to have a `bool` return value when this was exposed publicly, but now that it's internal, can we just have this return `void` instead?

abeltrano (changes requested):

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -480,6 +482,11 @@ UwbConnector::SetApplicationConfigurationParameters(uint32_t sessionId, std::vec
 void
 UwbConnector::HandleNotifications(std::stop_token stopToken)
 {
+    std::unique_lock lock{ m_callbacksPresentConditionVariableGate };
+    m_callbacksPresentConditionVariable.wait(lock, [this]() {
+        return CallbacksPresent();
+    });
```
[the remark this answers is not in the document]
forwardpointer: gotcha, will add

windows/devices/uwb/UwbDeviceConnector.cxx
forwardpointer: This function is actually not called by anything else, as its only role is to start the wait process so I think the constructor will have to call it. Another way, is actually to get rid of the condition variable and just have the callback registration codes grab the mutex, check if the were callbacks registered before, and if there were none then also start the notification listener. I think this method actually makes a lot of sense

windows/devices/uwb/UwbDeviceConnector.cxx
forwardpointer: If I understand correctly then scoped_lock behaves the same as lock_guard but also has support for grabbing multiple mutexes in succession in the same line. Since we're also considering combining the two mutexes, I think I'll stick with lock_guard. (Maybe there's less overhead)

windows/devices/uwb/UwbDeviceConnector.cxx
forwardpointer: #203

windows/devices/uwb/UwbDeviceConnector.cxx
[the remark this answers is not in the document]
forwardpointer: After merging with #203 this is outdated, unless we want to revert back to using the condition variable

windows/devices/uwb/include/windows/devices/uwb/UwbDeviceConnector.hxx
```diff
@@ -2,10 +2,13 @@
 #ifndef UWB_DEVICE_CONNECTOR_HXX
 #define UWB_DEVICE_CONNECTOR_HXX
 
+#include <condition_variable>
 #include <cstdint>
 #include <future>
 #include <memory>
+#include <mutex>
```
abeltrano, line 9: Is this still needed?

windows/devices/uwb/UwbDeviceConnector.cxx
abeltrano, line 33: Still doesn't seem like this is needed here. If it is, can you help me understand why?

corbin-phipps (dismissed):

windows/devices/uwb/UwbDeviceConnector.cxx
forwardpointer: Missed this, thanks

abeltrano (approved):

corbin-phipps (approved):

PR #200, opened 2023-04-03
Determine IOCTL_UWB_NOTIFICATION output buffer size using in-band size field

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Ensure the `UwbConnector` client processes `IOCTL_UWB_NOTIFICATION` output per the updated method which sets the required size in the `size` field and indicates `ERROR_INSUFFICIENT_BUFFER`. Update UWB simulator driver to mimic the UwbCx behavior.

### Technical Details

* Relax output buffer size requirement in uwb simulator driver to simply copy whatever data is available, and indicate `STATUS_BUFFER_OVERLFOW` if the output buffer wasn't at least as big as the output notification size.
* Update the notification processing code to read the bytes required from the partially filled in structure instead of from the `DeviceIoControl` output argument.

### Test Results

* Ran a full ranging scenario with the UWB simulator driver, enabling random measurements, and noted the correct state change values and measurements were delivered.

### Reviewer Focus

* I dislike this approach since it requires the caller have knowledge of the in-band data structure just to perform the i/o. This leaks information about the protocol into the transport, and these components should not normally have any knowledge of each other. This also means that i/o to the driver cannot be performed generically, resulting in some code bloat. We can live with it, but it's unfortunate.

### Future Work

* Other IOCTLs with variable-length output buffers need to be updated to follow the same pattern.

### Checklist


windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -514,6 +516,8 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
                     }
                 } else if (lastError == ERROR_MORE_DATA || lastError == ERROR_INSUFFICIENT_BUFFER) {
                     LOG_VERBOSE << "IOCTL_UWB_NOTIFICATION insufficient buffer, " << bytesRequired << " bytes required, current size " << std::size(uwbNotificationDataBuffer) << " bytes";
+                    const UWB_NOTIFICATION_DATA& notificationDataPartial = *reinterpret_cast<UWB_NOTIFICATION_DATA*>(std::data(uwbNotificationDataBuffer));
+                    bytesRequired = notificationDataPartial.size;
                     // Attempt to retry the ioctl with the appropriate buffer size, which is now held in bytesRequired.
                     continue;
                 } else {
```
abeltrano, line 523: This whole block can probably be removed as long as the driver always handles the request in an asynchronous manner. The simulator driver guarantees this, but I'm unsure of the UwbCx behavior.  Will re-visit this in another PR.

windows/devices/uwb/UwbDeviceConnector.cxx
```diff
@@ -502,6 +502,8 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
                         hr = HRESULT_FROM_WIN32(lastError);
                         if (lastError == ERROR_INSUFFICIENT_BUFFER || lastError == ERROR_MORE_DATA) {
                             // Driver indicated buffer was too small and put required size in 'bytesRequired'. Retry with new size.
+                            const UWB_NOTIFICATION_DATA& notificationDataPartial = *reinterpret_cast<UWB_NOTIFICATION_DATA*>(std::data(uwbNotificationDataBuffer));
```
corbin-phipps, line 505: If you wanted to avoid this cast, you could simply get the first 4 bytes of the buffer since that is where the size field is. Your way is probably safer, but this way would allow for some generalization (not in the notifications case, but for all of the other DDI structs). Since all of the DDI structs have 4-byte size fields in the beginning, this would allow us to avoid casting the buffer to the specific struct.

corbin-phipps (approved):

windows/devices/uwb/UwbDeviceConnector.cxx
abeltrano: I'm not willing to make any assumptions about a cross-process data structure, there's just too much risk. I have some other thoughts about the notification IOCTL that would affect this that I'll be bringing up this week. So, I'm going to wait and see how things shake out before generalizing the i/o with thread pools.

forwardpointer (approved):

corbin-phipps (approved):

windows/devices/uwb/UwbDeviceConnector.cxx
abeltrano: I came up with a few ideas for how to achieve generalized i/o with this method. Will share offline.

PR #201, opened 2023-04-04
Add input validation to remaining UCI app configuration parameters

corbin-phipps:
### Type

- [x] Bug fix
- [x] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR adds input validation to the remaining UCI app config parameters in nocli. Some bug fixes and minor improvements are also included.

### Technical Details

- Added missing fields to `PrfMode` and `RangeDataNotificationConfiguration` in `FiraDevice.hxx`.
- Updated `AddEnumOption` to display "RFrameConfiguration" instead of "StsPacketConfiguration".
- Added a function to do simple validation of all enum parameters.
- Added a function to do validation of all non-enum parameters and moved existing validation logic into it.
- Updated `ProcessApplicationConfigurationParameters` to handle all unsigned integers, not just uint8_t.
- Changed `appConfigParamsData` variable to `applicationConfigurationParametersData`.

### Test Results

Tested a few parameters of various types with invalid values - verified `std::cerr` output in nocli. Did not test every parameter.

### Reviewer Focus

There is a lot, partially due to the variable name change occurring in many places. The main focus should be on the two new validation functions.

Also, I don't like the name I gave to the new PrfMode parameter (`HprfSpecialDataRate`), but not sure what else to call it. It seems best to be more specific, but `HprfWithDataRate27_2And31_2Mbps` just doesn't seem right...

### Future Work

There are a lot of magic numbers that need to be changed. These are based on the UCI spec, so putting these in FiraDevice.hxx makes sense to me.

Right now, if a parameter is invalid, I just output a message to `std::cerr`. This is fine, but it might make more sense to either not send the invalid parameter and send all of the valid ones, or not send any parameters at all if one or more are invalid.

### Checklist

### DIFF
--- a/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
@@ -97,12 +97,14 @@ enum class StsConfiguration : uint8_t {
  * @brief See FiRa Consortium UWB MAC Technical Requirements v1.3.0, Section
  * 5.3.
  */
-enum class StsPacketConfiguration : uint8_t {
+enum class RFrameConfiguration : uint8_t {
     SP0 = 0,
     SP1 = 1,
     SP2 = 2,
     SP3 = 3,
 };
+
+using StsPacketConfiguration = RFrameConfiguration;
 
 /**
  * @brief See FiRa Consortium UWB MAC Technical Requirements v1.3.0, Section
@@ -190,13 +192,11 @@ enum class Channel {
 
 /**
  * @brief See FiRa Consortium UWB MAC Technical Requirements v1.3.0, Section
- * 7.5.3.3, Table 53; and FiRa Consortium UWB Command Interface Generic Technical
- * Specification v1.1.0, Section 8.3, Table 29
+ * 7.5.3.3, Table 53.
  */
 enum class PrfMode {
     Bprf = 0,
     Hprf = 1,
-    HprfSpecialDataRate = 2,
 };
 
 /**
@@ -547,7 +547,7 @@ using UwbApplicationConfigurationParameterValue = std::variant<
     KeyRotation, // KEY_ROTATION, tag 0x23, size 1
     MultiNodeMode, // MULTI_NODE_MODE, tag 0x03, size 1
     PreambleDuration, // PREAMBLE_DURATION, tag 0x17, size 1
-    PrfMode, // PRF_MODE, tag 0x1F, size 1
+    PrfModeDetailed, // PRF_MODE, tag 0x1F, size 1
     PsduDataRate, // PSDU_DATA_RATE, tag 0x16, size 1
     RangeDataNotificationConfiguration, // RANGE_DATA_NTF_CONFIG, tag 0x0E, size 1
     RangingRoundUsage, // RANGING_ROUND_USAGE, tag0x01, size 1
@@ -557,7 +557,7 @@ using UwbApplicationConfigurationParameterValue = std::variant<
     SchedulingMode, // SCHEDULED_MODE, tag 0x22, size 1
     StsConfiguration, // STS_CONFIG, tag 0x02, size 1
     StsLength, // STS_LENGTH, tag 0x035, length 1,
-    StsPacketConfiguration, // RFRAME_CONFIG, tag 0x12, size 1
+    RFrameConfiguration, // RFRAME_CONFIG, tag 0x12, size 1
     TxAdaptivePayloadPower, // TX_ADAPTIVE_PAYLOAD_POWER, tag 0x1C, size 1
     ::uwb::UwbMacAddress, // DEVICE_MAC_ADDRESS, tag 0x06, size 2/8
     ::uwb::UwbMacAddressFcsType, // MAC_FCS_TYPE, tag 0x0B, size 1
--- a/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
@@ -435,6 +435,7 @@ enum class AoAResult : uint8_t {
 enum class RangeDataNotificationConfiguration : uint8_t {
     Disable = 0U,
     Enable = 1U,
+    EnableInProximityRange = 2U,
 };
 
 enum class PreambleDuration : uint8_t {
--- a/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
@@ -190,11 +190,13 @@ enum class Channel {
 
 /**
  * @brief See FiRa Consortium UWB MAC Technical Requirements v1.3.0, Section
- * 7.5.3.3, Table 53.
+ * 7.5.3.3, Table 53; and FiRa Consortium UWB Command Interface Generic Technical
+ * Specification v1.1.0, Section 8.3, Table 29
  */
 enum class PrfMode {
     Bprf = 0,
     Hprf = 1,
+    HprfSpecialDataRate = 2,
 };
 
 /**
--- a/tools/cli/NearObjectCli.cxx
+++ b/tools/cli/NearObjectCli.cxx
@@ -233,14 +233,164 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory
 }
 
 /**
- * @brief Displays and returns application configuration parameters.
+ * @brief Validates input for nocli uwb range start.
+ * 
+ * @param cliData The nocli input data containing the application configuration parameters
+ */
+void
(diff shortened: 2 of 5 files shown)


lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
```diff
@@ -190,11 +190,13 @@ enum class Channel {
...
 /**
  * @brief See FiRa Consortium UWB MAC Technical Requirements v1.3.0, Section
- * 7.5.3.3, Table 53.
+ * 7.5.3.3, Table 53; and FiRa Consortium UWB Command Interface Generic Technical
+ * Specification v1.1.0, Section 8.3, Table 29
  */
 enum class PrfMode {
     Bprf = 0,
     Hprf = 1,
+    HprfSpecialDataRate = 2,
```
abeltrano, line 199: We already have the PRF define with rates in `PrfModeDetailed`, so is this actually needed here?

https://github.com/microsoft/nearobject-framework/blob/f916043c3f76aaa1afeb507af8775e9bbf22350b/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx#L450-L454

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
```diff
@@ -435,6 +437,7 @@ enum class AoAResult : uint8_t {
 enum class RangeDataNotificationConfiguration : uint8_t {
     Disable = 0U,
     Enable = 1U,
+    EnableInProximityRange = 2U,
```
abeltrano, line 440: Nice find, thanks!

tools/cli/NearObjectCli.cxx
```diff
@@ -211,7 +211,12 @@ CLI::Option*
 AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory = false)
 // clang-format on
 {
-    std::string optionName{ std::string("--").append(magic_enum::enum_type_name<EnumType>()) };
+    std::string optionName{ std::string("--") };
+    if (magic_enum::enum_type_name<EnumType>() == "StsPacketConfiguration") {
+        optionName.append("RFrameConfiguration");
```
abeltrano, line 216: Instead of this, is it possible to alias the enum? Eg. Add this to `FiraDevice.hxx` after the declaration of `enum class StsPacketConfiguration ... { `

`using RFrameConfiguration = StsPacketConfiguration;`

abeltrano:
> Also, I don't like the name I gave to the new PrfMode parameter (`HprfSpecialDataRate`), but not sure what else to call it. It seems best to be more specific, but `HprfWithDataRate27_2And31_2Mbps` just doesn't seem right...

We have a "detailed" enum that includes the rates in the field names. See the comment in the review for more details.

> 
> ### Future Work
> There are a lot of magic numbers that need to be changed. These are based on the UCI spec, so putting these in FiraDevice.hxx makes sense to me.

Putting them in `FiraDevice.hxx` as namespace `constexpr` variables with descriptive names would be great 👍

tools/cli/NearObjectCli.cxx
```diff
@@ -233,14 +238,200 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory
...
+            std::cerr << "Invalid ResultReportConfiguration length" << std::endl;
+            return false;
+        }
+        for (char c : resultReportConfigurationString) {
+            if (c != '0' && c != '1') {
+                std::cerr << "Invalid ResultReportConfiguration value" << std::endl;
+                return false;
+            }
+        }
+        return true;
```
abeltrano, line 363: Alternate way to do this with a little less code:
```suggestion
        return std::ranges::all_of(resultReportConfigurationString, [](const auto& c){ return c == '0' || c =='1'; });
```

tools/cli/NearObjectCli.cxx
```diff
@@ -233,14 +238,200 @@ AddEnumOption(CLI::App* app, std::optional<EnumType>& assignTo, bool isMandatory
...
+
+/**
+ * @brief Validates, displays, and returns application configuration parameters.
+ *
+ * @param cliData The nocli input data containing the application configuration parameters.
  * @return std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter>.
  */
 std::vector<uwb::protocol::fira::UwbApplicationConfigurationParameter>
-ProcessApplicationConfigurationParameters(UwbApplicationConfigurationParameterData applicationConfigurationParameterData)
+ProcessApplicationConfigurationParameters(std::shared_ptr<NearObjectCliData> cliData)
```
abeltrano, line 430: nit: This being a helper method and a completely synchronous call, you could safely pass a raw reference instead. Also, could a `const` ref be passed or does this function mutate `cliData`?

```suggestion
ProcessApplicationConfigurationParameters(const NearObjectCliData& cliData)
```

tools/cli/NearObjectCli.cxx
```diff
@@ -457,122 +649,76 @@ NearObjectCli::AddSubcommandUwbRawSessionDeinitialize(CLI::App* parent)
...
+    detail::AddEnumOption(rangeStartApp, applicationConfigurationParametersData.rFrameConfiguration);
+    detail::AddEnumOption(rangeStartApp, applicationConfigurationParametersData.scheduledMode);
+    detail::AddEnumOption(rangeStartApp, applicationConfigurationParametersData.stsConfiguration);
+    detail::AddEnumOption(rangeStartApp, applicationConfigurationParametersData.stsLength);
+    detail::AddEnumOption(rangeStartApp, applicationConfigurationParametersData.txAdaptivePayloadPower);
 
     // other
     rangeStartApp->add_option("--ResultReportConfig", m_cliData->resultReportConfigurationString, "4-bit report config, e.g. 0101. b3=AOA FOM, b2=AOA Elevation, b1=AOA Azimuth, b0=TOF")->capture_default_str();
-    rangeStartApp->add_option("--StaticStsInitializationVector", appConfigParamsData.staticStsIv, "6-octet array for vendor-defined static STS configuration, e.g. 11:22:33:44:55:66")->delimiter(':');
+    rangeStartApp->add_option("--StaticStsInitializationVector", applicationConfigurationParametersData.staticStsIv, "6-octet array for vendor-defined static STS configuration, e.g. 11:22:33:44:55:66")->delimiter(':');
```
abeltrano, line 718: nit: 
```suggestion
    rangeStartApp->add_option("--StaticStsInitializationVector", applicationConfigurationParametersData.staticStsIv, "6-octet array for vendor-defined static STS initialization vector, e.g. 11:22:33:44:55:66")->delimiter(':');
```

abeltrano (approved):
A few minor comments, but overall, this looks good. Thanks for taking on all the tedious typing!

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
corbin-phipps: Oh I missed that, I'll update my code!

tools/cli/NearObjectCli.cxx
corbin-phipps: It may be possible. I vaguely remember something in the docs for magic_enum that it doesn't work with aliasing, but I will give it a try!

tools/cli/NearObjectCli.cxx
corbin-phipps: Doesn't mutate `cliData`, so I will change this to pass a raw const ref

tools/cli/NearObjectCli.cxx
abeltrano: The alternative would be to rename the enum, then introduce an alias for the current name so that it can continue to be used as-is, but you can use `RFrameConfiguration` in the CLI.

corbin-phipps: Yeah unfortunately I can confirm that magic_enum doesn't recognize the alias, so I will try your second suggestion - rename the enum to `RFrameConfiguration` and add an alias like `using StsPacketConfiguration = RFrameConfiguration`

corbin-phipps:
> > Also, I don't like the name I gave to the new PrfMode parameter (`HprfSpecialDataRate`), but not sure what else to call it. It seems best to be more specific, but `HprfWithDataRate27_2And31_2Mbps` just doesn't seem right...
> 
> We have a "detailed" enum that includes the rates in the field names. See the comment in the review for more details.
> 
> > ### Future Work
> > There are a lot of magic numbers that need to be changed. These are based on the UCI spec, so putting these in FiraDevice.hxx makes sense to me.
> 
> Putting them in `FiraDevice.hxx` as namespace `constexpr` variables with descriptive names would be great 👍

Sounds good, I will do this in a separate PR since this one is getting long

abeltrano (approved):

PR #202, opened 2023-04-04
Add WinRT component project for Windows service WinRT API

abeltrano:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Allow development of Windows service APIs.

### Technical Details

* Add external Windows Runtime Component project, `nearobject-windows-winrt` to CMake project tree.

### Test Results

* Project builds both in CMake and directly using the MSBuild solution file.
* Add skeleton runtime classes for the 2 top-level objects, `NearObjectSession` and `NearObject`.

### Reviewer Focus

None

### Future Work

* This project may need a separate `.clang-format` file since the project uses C++/WinRT which will probably necessitate a slightly different coding style.

### Checklist


windows/nearobject/winrt/windows.devices.nearobject/pch.cpp
```diff
@@ -0,0 +1 @@
+﻿#include "pch.h"
```
corbin-phipps, line 1: IIRC, you're not a big fan of precompiled headers. Is this something that is required by winrt?

corbin-phipps (dismissed):

windows/nearobject/winrt/windows.devices.nearobject/pch.cpp
abeltrano: I figured out how to do disable use of PCH in `cppwinrt.exe`; there's a switch `-pch <pch filename>` that can be specified, and using `'.'` as the value disables PCH entirely. In doing so, I figured out we absolutely should not disable PCH in the generated projection bindings 😄

The generated WinRT headers (`winrt/Windows.Foundation.h, winrt/Windows.Foundation.Collections.h`) are massive and pretty complex. Some of them are several MBs in size. This is exactly the situation where PCH provides benefit since we don't want the compiler processing several MB headers for each compilation unit. So, I think we should keep the PCH and document that it should be used sparingly, possibly only for C++/WinRT projection bindings.

corbin-phipps (approved):

windows/nearobject/winrt/windows.devices.nearobject/pch.cpp
abeltrano: Documented the restrictions on PCH use.

PR #213, opened 2023-04-06
Drop std::optional from GetSessionCount() return value type

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Reduce ambiguity in `GetSessionCount()` return values (`!has_value()` and `.value() == 0` are ambiguous).

### Technical Details

* Use `uint32_t directly instead of `std::optional<uint32_t>` since the latter is unnecessary.

### Test Results

Compile-tested only.

### Reviewer Focus

None

### Future Work

None

### Checklist

### DIFF
--- a/lib/uwb/UwbDevice.cxx
+++ b/lib/uwb/UwbDevice.cxx
@@ -149,7 +149,7 @@ UwbDevice::GetDeviceInformation()
     return GetDeviceInformationImpl();
 }
 
-std::optional<uint32_t>
+uint32_t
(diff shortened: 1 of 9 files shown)


corbin-phipps (approved):
Thanks for this, I agree that std::optional is unnecessary here

Issue #219, opened 2023-04-07
Bug: GetApplicationConfigurationParameters does not follow FiRa UCI Generic Technical Spec v1.1.0


corbin-phipps:
**Describe the bug**
The UCI specification states that when a client requests the application configuration parameters from a device (through `GetApplicationConfigurationParameters`), the client should either:
1. Supply a list of requested parameter types and the number of requested parameters, in which case the device will either return the requested parameters if they were set in the device or will return all of the invalid parameters (and no valid parameters), as well as `STATUS_INVALID_PARAMETER` if any requested parameter type does not exist in the device.
2. Supply a "0" for the number of requested parameters and no list of parameter types, in which case the device will return all of the parameters that have been set in the device.

**Expected behavior**
When the client (nocli for example) wants all of the currently set parameters in the device (aka the current implementation of `GetApplicationConfigurationParameters`), the client should not supply a list of parameter types and should set the number of parameters to 0.

**Actual behavior**
Whenever the nocli client sets any number of app config parameters (besides all of them), nocli then requests the set parameters by currently sending a list of all parameter types, resulting in `STATUS_INVALID_PARAMETER` and a long list of invalid parameters when testing with the TestClient driver.

**Screenshots**
<img width="513" alt="image" src="https://user-images.githubusercontent.com/43283510/230645322-eb585745-c3ab-44c9-a627-7d17b7d01194.png">

PR #220, opened 2023-04-07
Update UwbConnector + simulator driver to handle in-band output buffer size indication

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR updates the IOCTL invocation code to use the proper in-band indication for the required output buffer size. It also updates the simulator driver's dispatch table accordingly, matching the dispatch table in the CX driver.

### Technical Details

- Use `.size` field for resizing output buffers instead of what was previously returned in `bytesRequired`.
- Rewrite `GetCapabilities()` to match format and style of other similar functions.
- Update Simulator driver dispatch table.

### Test Results

Mostly compile tested for now. Can confirm that the new changes now allows GetApplicationConfigurationParameters to communicate with the TestClient driver (although #219 still causes failures)

### Reviewer Focus

Does the Simulator driver need additional changes besides the dispatch table updates?

### Future Work

Adding a raw command `GetDeviceCapabilities` could allow us to test that functionality. Also, additional testing could be done for the new resizing logic with variable size arrays (e.g. I tested GetDeviceInfo with the TestClient driver, but that currently doesn't set any vendor specific info, so the resizing logic wasn't actually tested).

### Checklist

### DIFF
--- a/windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
+++ b/windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
@@ -45,11 +45,11 @@ UwbSimulatorDispatchEntryLrp (*MakeLrpDispatchEntry)(ULONG, typename UwbSimulato
 const std::initializer_list<UwbSimulatorDispatchEntry<UwbSimulatorDdiHandler>> UwbSimulatorDdiHandler::Dispatch{
     // GUID_UWB_DEVICE_INTERFACE Handlers
     MakeLrpDispatchEntry<UWB_DEVICE_RESET, UWB_STATUS>(IOCTL_UWB_DEVICE_RESET, &UwbSimulatorDdiHandler::OnUwbDeviceReset),
-    MakeLrpDispatchEntry<Unrestricted, Unrestricted>(IOCTL_UWB_GET_DEVICE_INFO, &UwbSimulatorDdiHandler::OnUwbGetDeviceInformation),
-    MakeLrpDispatchEntry<Unrestricted, Unrestricted>(IOCTL_UWB_GET_DEVICE_CAPABILITIES, &UwbSimulatorDdiHandler::OnUwbGetDeviceCapabilities),
+    MakeLrpDispatchEntryWithSize(IOCTL_UWB_GET_DEVICE_INFO, &UwbSimulatorDdiHandler::OnUwbGetDeviceInformation, 0, offsetof(UWB_DEVICE_INFO, vendorSpecificInfo)),
+    MakeLrpDispatchEntryWithSize(IOCTL_UWB_GET_DEVICE_CAPABILITIES, &UwbSimulatorDdiHandler::OnUwbGetDeviceCapabilities, 0, offsetof(UWB_DEVICE_CAPABILITIES, capabilityParams)),
     MakeLrpDispatchEntry<UWB_SET_DEVICE_CONFIG_PARAMS, Unrestricted>(IOCTL_UWB_GET_DEVICE_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbGetDeviceConfigurationParameters),
     MakeLrpDispatchEntry<UWB_GET_DEVICE_CONFIG_PARAMS, Unrestricted>(IOCTL_UWB_SET_DEVICE_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbSetDeviceConfigurationParameters),
-    MakeLrpDispatchEntry<UWB_GET_APP_CONFIG_PARAMS, Unrestricted>(IOCTL_UWB_GET_APP_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbGetApplicationConfigurationParameters),
+    MakeLrpDispatchEntryWithSize(IOCTL_UWB_GET_APP_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbGetApplicationConfigurationParameters, offsetof(UWB_GET_APP_CONFIG_PARAMS, appConfigParams), offsetof(UWB_APP_CONFIG_PARAMS, appConfigParams)),
     MakeLrpDispatchEntryWithSize(IOCTL_UWB_SET_APP_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbSetApplicationConfigurationParameters, offsetof(UWB_SET_APP_CONFIG_PARAMS, appConfigParams), offsetof(UWB_SET_APP_CONFIG_PARAMS_STATUS, appConfigParamsStatus)),
     MakeLrpDispatchEntry<Unrestricted, UWB_GET_SESSION_COUNT>(IOCTL_UWB_GET_SESSION_COUNT, &UwbSimulatorDdiHandler::OnUwbGetSessionCount),
     MakeLrpDispatchEntry<UWB_SESSION_INIT, UWB_STATUS>(IOCTL_UWB_SESSION_INIT, &UwbSimulatorDdiHandler::OnUwbSessionInitialize),
--- a/windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
+++ b/windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
@@ -103,12 +103,15 @@ UwbSimulatorDdiHandler::OnUwbGetDeviceInformation(WDFREQUEST request, std::span<
 
     // Convert neutral type to DDI output type.
     auto uwbDeviceInformation = UwbCxDdi::From(deviceInformation);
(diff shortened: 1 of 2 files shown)


forwardpointer (dismissed):

windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
```diff
@@ -45,11 +45,11 @@ UwbSimulatorDispatchEntryLrp (*MakeLrpDispatchEntry)(ULONG, typename UwbSimulato
...
+    MakeLrpDispatchEntryWithSize(IOCTL_UWB_GET_DEVICE_INFO, &UwbSimulatorDdiHandler::OnUwbGetDeviceInformation, 0, offsetof(UWB_DEVICE_INFO, vendorSpecificInfo)),
+    MakeLrpDispatchEntryWithSize(IOCTL_UWB_GET_DEVICE_CAPABILITIES, &UwbSimulatorDdiHandler::OnUwbGetDeviceCapabilities, 0, offsetof(UWB_DEVICE_CAPABILITIES, capabilityParams)),
     MakeLrpDispatchEntry<UWB_SET_DEVICE_CONFIG_PARAMS, Unrestricted>(IOCTL_UWB_GET_DEVICE_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbGetDeviceConfigurationParameters),
     MakeLrpDispatchEntry<UWB_GET_DEVICE_CONFIG_PARAMS, Unrestricted>(IOCTL_UWB_SET_DEVICE_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbSetDeviceConfigurationParameters),
-    MakeLrpDispatchEntry<UWB_GET_APP_CONFIG_PARAMS, Unrestricted>(IOCTL_UWB_GET_APP_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbGetApplicationConfigurationParameters),
+    MakeLrpDispatchEntryWithSize(IOCTL_UWB_GET_APP_CONFIG_PARAMS, &UwbSimulatorDdiHandler::OnUwbGetApplicationConfigurationParameters, offsetof(UWB_GET_APP_CONFIG_PARAMS, appConfigParams), offsetof(UWB_APP_CONFIG_PARAMS, appConfigParams)),
```
abeltrano, line 52: These will require changes to the handler functions (listed in each dispatch entry here) to provide partial responses. Right now, they'll only populate the output buffer if the response is large enough. Shouldn't be too much work. The handler for `IOCTL_UWB_SET_APP_CONFIG_PARAMS` (`OnUwbSetApplicationConfigurationParameters`) does this correctly, so you can use that as a template.

abeltrano: https://github.com/microsoft/nearobject-framework/blob/f10adf9f05e2b242f594087c29b9cf35c9a22657/windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx#L258-L275

abeltrano (changes requested):

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -149,33 +151,40 @@ UwbConnector::GetCapabilities()
...
+            // Treat all errors other than insufficient buffer size as fatal.
+            if (lastError != ERROR_MORE_DATA) {
+                HRESULT hr = HRESULT_FROM_WIN32(lastError);
+                PLOG_ERROR << "error when sending IOCTL_UWB_GET_DEVICE_CAPABILITIES, hr=" << std::showbase << std::hex << hr;
+                resultPromise.set_exception(std::make_exception_ptr(UwbException(UwbStatusGeneric::Failed)));
+                break;
+            }
+            // Attempt to retry the ioctl with the appropriate buffer size, which is now held in bytesRequired.
+            const auto& deviceCapabilitiesPartial = *reinterpret_cast<UWB_DEVICE_CAPABILITIES*>(std::data(uwbDeviceCapabilitiesBuffer));
+            bytesRequired = deviceCapabilitiesPartial.size;
```
abeltrano, line 172: One thing I forgot to do in the pre-existing impl was to validate that `bytesRequired >= offsetof(UWB_DEVICE_CAPABILITIES, size)`. Without that, we'd be accessing invalid memory if the driver didn't return enough data. Not likely to happen, but we need to code defensively against it.

Can you please add that here and to the existing ones? (I think `IOCTL_UWB_NOTIFICATION` and `IOCTL_UWB_SET_APP_CONFIG_PARAMS`).

windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
corbin-phipps: I updated the handlers to follow the `OnUwbSetApplicationConfigurationParameters` format but did not test them, so this may require a closer look

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: I don't quite understand the need for this. If the `size` field is later in the struct, how would that affect the buffer size? It's the same value right?

abeltrano: Here, the amount of data returned by the ioctl is `bytesRequired`. So, before we access stuff inside `deviceCapabilitiesPartial`, we need to ensure that we're accessing a part that's actually been filled in by the ioctl. Since we're accessing the `size` field, we know it's at offset `offsetof(UWB_DEVICE_CAPABILITIES, size)` and it has size `sizeof(UWB_DEVICE_CAPABILITIES::size)`, so, `bytesRequired` has to be >= `offsetof(UWB_DEVICE_CAPABILITIES, size) + offsetof(UWB_DEVICE_CAPABILITIES, size)`, otherwise we'd be accessing memory past the end of the ioctl output buffer. My main point was that using this expression for calculating the validity of accessing the `size` member doesn't make any assumptions about where `size` is in the partially filled structure. Does that make more sense?

corbin-phipps: I see, I understand the reasoning now. However, it's not clear where this check would happen. The amount of data returned by the IOCTL isn't in `bytesRequired`, we set `bytesRequired = deviceInformationPartial.size`, so the check would need to happen after this. But this still results in the same problem in the case where the `size` field is at the end of the struct and the buffer is too small to contain that field - we'd be accessing invalid memory when reading `deviceInformationPartial.size`. Right?

corbin-phipps: I see, I understand the reasoning now. However, it's not clear where this check would happen. The amount of data returned by the IOCTL isn't in `bytesRequired` right after the `DeviceIoControl` call, the amount of data would go in the `size` field, then we set `bytesRequired = deviceInformationPartial.size`. So the check would need to happen after this, but this still results in the same problem in the case where the `size` field is at the end of the struct and the buffer is too small to contain that field - we'd be accessing invalid memory when reading `deviceInformationPartial.size`. Right?

windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
abeltrano: Let's test them please so we know whether the updated code works.

windows/devices/uwb/UwbConnector.cxx
abeltrano: You're right, yeah. This is a dilemma since this approach is inconsistent with the `DeviceIoControl` contract which states this:

image

That's exactly the case here: `ERROR_GET_MORE_DATA` was returned and the `DeviceIoContrtol` API tells us that the amount of data returned is in `lpBytesReturned`, which is what one would expect and is intuitive.

Let's leave it as-is for now. I'd like to think about it some more and see if there's a way to make this consistent, intuitive, and doesn't break the `DeviceIoControl` contract.

windows/drivers/uwb/simulator/UwbSimulatorDdiHandler.cxx
corbin-phipps: I can confirm that the changes in `OnUwbGetApplicationConfigurationParameters` work, which I would consider to be the "most complicated" example out of the 3 changes. 

The current functionality in `OnUwbGetDeviceInformation` still works, although we currently don't test the variable size array here (vendor specific info).

We don't have a raw command for device capabilities, so that one is untested.

abeltrano: Thanks for doing the tests. We don't need to worry about the ones that aren't exercised (ie. those without raw commands from the cli), so no worries there.

If you can confirm `OnUwbGetApplicationConfigurationParameters` works with the simulator driver, then we're good to go.

PR #224, opened 2023-04-11
Callback Deregistration

forwardpointer:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [X] Infrastructure

### Goals

Allow the deregistration of callbacks

### Technical Details

* Every callback has an id, this map is stored in m_[session|device]EventCallbacksIdMap. 
* m_[session|device]EventCallbacks now stores the callback ids, not the callbacks themselves.
* m_tokens holds the list of tokens so that the memory is not released when we return the tokens to the caller of the register callback functions

### Test Results

* the basic ranging scenario works

### Reviewer Focus

Currently the interface maintains the invariant that m_tokens, m_[session|device]EventCallbacks, and m_[session|device]EventCallbacksIdMap are always in sync. However, this may result in a slow runtime for GetResolved[Session|Device]EventCallbacks(). Consider if relaxing that invariant makes sense.

### Checklist

### DIFF
--- a/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
+++ b/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
@@ -17,6 +17,10 @@ namespace uwb
  */
 class RegisteredCallbackToken;
 
+class RegisteredSessionCallbackToken;
+
+class RegisteredDeviceCallbackToken;
+
 /**
  * @brief Interface for receiving events from a UwbSession. This is the primary
  * method to receive information from near peers.
--- a/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
+++ b/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
@@ -17,10 +17,6 @@ namespace uwb
  */
 class RegisteredCallbackToken;
 
-class RegisteredSessionCallbackToken;
-
-class RegisteredDeviceCallbackToken;
-
 /**
  * @brief Interface for receiving events from a UwbSession. This is the primary
  * method to receive information from near peers.
--- a/windows/devices/uwb/UwbConnector.cxx
+++ b/windows/devices/uwb/UwbConnector.cxx
@@ -934,18 +934,9 @@ UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackTok
     }
     std::lock_guard eventCallbacksLockExclusive{ m_eventCallbacksGate };
 
-    auto deviceCallback = std::dynamic_pointer_cast<::uwb::RegisteredDeviceCallbackToken>(tokenShared);
-    if (deviceCallback) {
-        // treat it as a device callback
-        auto tokenIt = std::find_if(std::cbegin(m_deviceEventCallbacks), std::cend(m_deviceEventCallbacks), [deviceCallback](const auto& token) {
-            return token.get() == deviceCallback.get();
-        });
-        if (tokenIt == std::cend(m_deviceEventCallbacks)) {
-            return; // no associated token found, bail
-        }
-        m_deviceEventCallbacks.erase(tokenIt);
-    } else {
-        auto sessionCallback = dynamic_pointer_cast<::uwb::RegisteredSessionCallbackToken>(tokenShared);
+    auto sessionCallback = dynamic_pointer_cast<::uwb::RegisteredSessionCallbackToken>(tokenShared);
+    if (sessionCallback) {
+        // treat it as a session callback
         auto sessionId = sessionCallback->SessionId;
 
         auto node = m_sessionEventCallbacks.extract(sessionId);
@@ -962,6 +953,18 @@ UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackTok
         }
         tokens.erase(tokenIt);
         m_sessionEventCallbacks.insert(std::move(node));
+    } else {
+        auto deviceCallback = std::dynamic_pointer_cast<::uwb::RegisteredDeviceCallbackToken>(tokenShared);
+        if (deviceCallback == nullptr) {
+            throw std::runtime_error("invalid callback type used, this is a bug!");
+        }
+        auto tokenIt = std::find_if(std::cbegin(m_deviceEventCallbacks), std::cend(m_deviceEventCallbacks), [deviceCallback](const auto& token) {
+            return token.get() == deviceCallback.get();
+        });
+        if (tokenIt == std::cend(m_deviceEventCallbacks)) {
+            return; // no associated token found, bail
+        }
+        m_deviceEventCallbacks.erase(tokenIt);
     }
 }
 
--- a/windows/devices/uwb/UwbConnector.cxx
+++ b/windows/devices/uwb/UwbConnector.cxx
@@ -23,21 +23,21 @@ namespace uwb
 {
 struct RegisteredCallbackToken
 {
-    virtual ~RegisteredCallbackToken(){};
+    virtual ~RegisteredCallbackToken() = default;
 };
 
 struct RegisteredSessionCallbackToken : public RegisteredCallbackToken
 {
-    RegisteredSessionCallbackToken(uint32_t sessionId, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> Callbacks) :
-        sessionId(sessionId),
-        Callbacks(Callbacks){};
-    uint32_t sessionId;
+    RegisteredSessionCallbackToken(uint32_t sessionId, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> callbacks) :
+        SessionId(sessionId),
+        Callbacks(std::move(callbacks)){};
+    uint32_t SessionId;
     std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> Callbacks;
 };
 
 struct RegisteredDeviceCallbackToken : public RegisteredCallbackToken
 {
-    RegisteredDeviceCallbackToken(std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> Callbacks) :
+    RegisteredDeviceCallbackToken(std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> callbacks) :
         Callbacks(Callbacks){};
     std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> Callbacks;
 };
@@ -946,7 +946,7 @@ UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackTok
         m_deviceEventCallbacks.erase(tokenIt);
     } else {
         auto sessionCallback = dynamic_pointer_cast<::uwb::RegisteredSessionCallbackToken>(tokenShared);
-        auto sessionId = sessionCallback->sessionId;
+        auto sessionId = sessionCallback->SessionId;
 
         auto node = m_sessionEventCallbacks.extract(sessionId);
         if (node.empty()) {
--- a/windows/devices/uwb/UwbConnector.cxx
+++ b/windows/devices/uwb/UwbConnector.cxx
@@ -919,20 +919,14 @@ UwbConnector::CallbacksPresent()
 void
 UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackToken> token)
 {
-    auto tok = token.lock();
-    if (not tok) {
+    auto tokenShared = token.lock();
+    if (not tokenShared) {
         return;
     }
     std::lock_guard eventCallbacksLockExclusive{ m_eventCallbacksGate };
 
-    DeregisterEventCallback(*tok);
-}
-
-void
-UwbConnector::DeregisterEventCallback(::uwb::RegisteredCallbackToken token)
-{
-    auto callbackId = token.callbackId;
-    auto isDeviceEventCallback = token.isDeviceEventCallback;
+    auto callbackId = tokenShared->callbackId;
+    auto isDeviceEventCallback = tokenShared->isDeviceEventCallback;
 
     // remove the token from the list of tokens
     for (auto it = std::cbegin(m_tokens); it != std::cend(m_tokens); it++) {
@@ -946,6 +940,7 @@ UwbConnector::DeregisterEventCallback(::uwb::RegisteredCallbackToken token)
         // first remove the id from the id map
         auto result = m_deviceEventCallbacksIdMap.find(callbackId);
         if (result == std::cend(m_deviceEventCallbacksIdMap)) {
+            PLOG_INFO << "the token is stale, the id is not present in the map. callbackId=" << callbackId;
             return;
         }
         m_deviceEventCallbacksIdMap.erase(result);
@@ -953,6 +948,7 @@ UwbConnector::DeregisterEventCallback(::uwb::RegisteredCallbackToken token)
         // first remove the id from the id map
         auto result = m_sessionEventCallbacksIdMap.find(callbackId);
         if (result == std::cend(m_sessionEventCallbacksIdMap)) {
+            PLOG_INFO << "the token is stale, the id is not present in the map. callbackId=" << callbackId;
             return;
         }
         m_sessionEventCallbacksIdMap.erase(result);
@@ -973,7 +969,7 @@ UwbConnector::DeregisterEventCallback(::uwb::RegisteredCallbackToken token)
             auto& id = *it;
             if (id == callbackId) {
                 callbackIds.erase(it);
-                return;
+                break;
             }
         }
 
--- a/windows/devices/uwb/UwbConnector.cxx
+++ b/windows/devices/uwb/UwbConnector.cxx
@@ -900,7 +900,8 @@ UwbConnector::RegisterDeviceEventCallbacks(std::weak_ptr<::uwb::UwbRegisteredDev
 {
     std::lock_guard eventCallbacksLockExclusive{ m_eventCallbacksGate };
     bool isFirstCallback = not CallbacksPresent();
-    auto token = InsertDeviceEventCallback(callbacks);
+    auto token = std::make_shared<::uwb::RegisteredDeviceCallbackToken>(callbacks);
+    m_deviceEventCallbacks.push_back(token);
     if (isFirstCallback) {
         NotificationListenerStart();
     }
@@ -912,7 +913,15 @@ UwbConnector::RegisterSessionEventCallbacks(uint32_t sessionId, std::weak_ptr<::
 {
     std::lock_guard eventCallbacksLockExclusive{ m_eventCallbacksGate };
     bool isFirstCallback = not CallbacksPresent();
-    auto token = InsertSessionEventCallback(sessionId, callbacks);
+    auto token = std::make_shared<::uwb::RegisteredSessionCallbackToken>(sessionId, callbacks);
+    auto node = m_sessionEventCallbacks.extract(sessionId);
+    if (!node.empty()) {
+        auto& sessionEventCallbacks = node.mapped();
+        sessionEventCallbacks.push_back(token);
+        m_sessionEventCallbacks.insert(std::move(node));
+    } else {
+        m_sessionEventCallbacks.insert({ sessionId, { token } });
+    }
     if (isFirstCallback) {
         NotificationListenerStart();
     }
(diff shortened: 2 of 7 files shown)


windows/devices/uwb/UwbConnector.cxx
```diff
@@ -915,8 +939,90 @@ UwbConnector::CallbacksPresent()
...
 }
+
+void
+UwbConnector::DeregisterEventCallback(::uwb::RegisteredCallbackToken token)
+{
+    auto callbackId = token.callbackId;
+    auto isDeviceEventCallback = token.isDeviceEventCallback;
+
+    // first remove the token from the list of tokens
+    for (auto it = std::cbegin(m_tokens); it != std::cend(m_tokens); it++) {
```
corbin-phipps, line 960: You could use `std::erase_if` here instead. For example,

std::erase_if(m_tokens, [&callbackId](const auto& token) {
    return token->callbackId == callbackId;
});

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -915,8 +939,90 @@ UwbConnector::CallbacksPresent()
...
+        // first remove the id from the id map
+        auto result = m_sessionEventCallbacksIdMap.find(callbackId);
+        if (result == std::cend(m_sessionEventCallbacksIdMap)) {
+            return;
+        }
+        m_sessionEventCallbacksIdMap.erase(result);
+
+        // now remove the id from the list of callbacks
+        for (auto sessionIt = std::cbegin(m_sessionEventCallbacks); sessionIt != std::cend(m_sessionEventCallbacks); sessionIt++) {
+            auto [sessionId, sessionCallbackIds] = *sessionIt;
```
corbin-phipps, line 993: You could use the structured binding directly in the for loop:

`for (const auto& [sessionId, sessionCallbackIds] : m_sessionEventCallbacks)`

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
```diff
@@ -226,15 +221,52 @@ private:
...
+    void
+    DeregisterEventCallback(::uwb::RegisteredCallbackToken token);
+
 private:
     // the following shared_mutex is used to protect access to both session and device event callbacks
     mutable std::shared_mutex m_eventCallbacksGate;
-    std::unordered_map<uint32_t, std::vector<std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks>>> m_sessionEventCallbacks;
-    std::vector<std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks>> m_deviceEventCallbacks;
+    std::unordered_map<uint32_t, std::vector<uint32_t>> m_sessionEventCallbacks;
+    std::vector<uint32_t> m_deviceEventCallbacks;
```
corbin-phipps, line 258: It might be useful to rename this to `m_deviceEventCallbackIds` since this no longer stores the callbacks themselves

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
```diff
@@ -226,15 +221,52 @@ private:
...
     std::string m_deviceName{};
     std::jthread m_notificationThread;
     wil::shared_hfile m_notificationHandleDriver;
     OVERLAPPED m_notificationOverlapped;
+
+private:
+    // the following is related to how the UwbConnector keeps track of its callbacks
+    uint32_t m_tokenUniqueState;
+    std::unordered_map<uint32_t, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks>> m_sessionEventCallbacksIdMap;
```
corbin-phipps, line 267: I think the names may be switched from the members above. This is called `m_sessionEventCallbacksIdMap` but stores the actual callbacks, while above you have `m_sessionEventCallbacks` which is a map that stores the callback IDs.

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
```diff
@@ -226,15 +221,52 @@ private:
...
+     */
+    std::shared_ptr<::uwb::RegisteredCallbackToken>
+    InsertDeviceEventCallback(std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> callback);
+
+    /**
+     * @brief Internal helper function to insert the callback into the hashmap of ids to callbacks
+     * You MUST have grabbed the m_eventCallbacksGate mutex prior to calling this.
+     *
+     * @param callback
+     * @return std::weak_ptr<::uwb::RegisteredCallbackToken>
```
corbin-phipps, line 239: The function returns `std::shared_ptr`, so the comment should be updated

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
```diff
@@ -226,15 +221,52 @@ private:
     std::vector<std::shared_ptr<::uwb::UwbRegisteredDeviceEventCallbacks>>
     GetResolvedDeviceEventCallbacks();
 
+    /**
+     * @brief Internal helper function to insert the callback into the hashmap of ids to callbacks
+     * You MUST have grabbed the m_eventCallbacksGate mutex prior to calling this.
+     *
+     * @param callback
+     * @return std::weak_ptr<::uwb::RegisteredCallbackToken>
```
corbin-phipps, line 229: The function returns `std::shared_ptr`, so this comment should be updated

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -915,8 +939,90 @@ UwbConnector::CallbacksPresent()
...
+                }
+            }
+        }
+    }
+}
+
+std::shared_ptr<::uwb::RegisteredCallbackToken>
+UwbConnector::InsertDeviceEventCallback(std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> callback)
+{
+    m_tokenUniqueState++;
```
corbin-phipps, line 1015: This isn't clear. What is the unique state and why is it incremented as an integer?

corbin-phipps (changes requested):

windows/devices/uwb/UwbConnector.cxx
forwardpointer: the for loop also erases sessionIt from m_sessionEventCallbacks if it becomes empty, so this won't be possible

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
forwardpointer: m_sessionEventCallbacksIdMap stores the map of the id to the actual callback. I agree that m_sessionEventCallbacks is a misnomer, so I'll rename it

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
forwardpointer: not sure what you're referring to here, as the function returns a weak_ptr

windows/devices/uwb/UwbConnector.cxx
forwardpointer: m_tokenUniqueState is just a (admittedly, hacky) way to maintain unique ids for each callback that gets registered. Each time a callback is registered, it gets assigned an id of m_tokenUniqueState and then m_tokenUniqueState is incremented.

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -633,17 +634,28 @@ UwbConnector::GetResolvedDeviceEventCallbacks()
...
+            return {};
+        }
+        auto& [id, deviceEventCallbackWeak] = *result;
         auto deviceEventCallback = deviceEventCallbackWeak.lock();
         if (deviceEventCallback != nullptr) {
             deviceEventCallbacks.push_back(std::move(deviceEventCallback));
-            it = std::next(it);
         } else {
-            it = m_deviceEventCallbacks.erase(it);
+            staleCallbacks.push_back(callbackId);
```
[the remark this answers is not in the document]
forwardpointer: That function changes the container that the for loop is iterating over, so it might not be possible to do so. (I'm unsure of exactly how the erase() function invalidates subsequent iterators)

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: I see. In that case, I think the name is misleading since the variable is meant to track a unique ID number but the name implies that it's tracking some sort of state.

windows/devices/uwb/UwbConnector.cxx
abeltrano: +1, std:;erase_if() is is safer. `erase()` on its own invalidates the iterator, so calling `it++` here would be accessing invalid memory. `erase()` returns the iterator following the erased element, so the idiomatic way to erase is to do this:

```
it = containerItCameFrom.erase(it);
```

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
abeltrano: Seems the return type should be `std::weak_ptr<::uwb::RegisteredCallbackToken>` since the public API function (correctly) returns as-such.

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
abeltrano: Like other function, the return type should just be changed to `std::weak_ptr<::uwb::RegisteredCallbackToken>`.

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
```diff
@@ -226,15 +221,52 @@ private:
...
     std::jthread m_notificationThread;
     wil::shared_hfile m_notificationHandleDriver;
     OVERLAPPED m_notificationOverlapped;
+
+private:
+    // the following is related to how the UwbConnector keeps track of its callbacks
+    uint32_t m_tokenUniqueState;
+    std::unordered_map<uint32_t, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks>> m_sessionEventCallbacksIdMap;
+    std::unordered_map<uint32_t, std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks>> m_deviceEventCallbacksIdMap;
+    std::vector<std::shared_ptr<::uwb::RegisteredCallbackToken>> m_tokens;
```
abeltrano, line 269: There's a lot of state here and having multiple maps makes it complicated to track. A few observations:

* `m_uniqueTokenState` is used to uniquely identify a specific callback registration. For each such registration, we're creating a `std::shared_ptr<RegisteredCallbackToken>`. So, we could do at least one of two things here to reduce the complexity and consolidate: 
    - Move the uniqueness into `RegisteredCallbackToken` directly, where it has a `private static` function to generate a unique identifier and assigns it to `callbackId` upon construction. This could just use a `static std::atomic<uint32_>` to generate new unique identifiers.
    - Remove `callbackId` all-together and instead use the fact that the memory address of the `shared_ptr<RegisteredCallbackToken>` (ie. that obtained from `.get()`) is itself unique and could be used to look up the token. Eg.

```
void UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackToken> tokenWeak) {
    auto tokenToDeregister = tokenWeak.lock():
    auto tokenIt = std::find_if(m_tokens, [](const auto& token) { return token.get() == tokenToDeregister.get(); });
    if (tokenIt == std::cend(m_tokens)) {
        return; // no associated token found, bail
    }
    std::shared_ptr<::uwb::RegisteredCallbackToken> token = *tokenIt;
    // do whatever is needed with token
    // erase token from token container
    m_tokens.erase(tokenIt);
}
```

* It seems like we could store the callbacks themselves in `::uwb::RegisteredCallbackToken` since we know we're going to have one for each registration. So why not consolidate? The `RegisteredCallbackToken` then becomes more like a callback state tracking structure rather than just the token. Eg.

```
struct RegisteredCallbackToken;
struct DeviceRegisteredCallbackToken : public RegisteredCallbackToken {
    std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> Callbacks;
};
struct SessionRegisteredCallbackToken : public RegisteredCallbackToken {
    std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> Callbacks;
};

std::vector<std::shared_ptr<DeviceRegisteredCallbackToken>> m_deviceCallbacks;
std::unordered_map<uint32_t, std::shared_ptr<SessionRegisteredCallbackToken >> m_sessionEventCallbacks;
```

When you get a `std::weak_ptr<RegisteredCallbackToken>`, you'd need to use `std::dynamic_cast<T>` to determine if you've been passed a `DeviceRegisteredCallbackToken` or a `SessionRegisteredCallbackToken`.  

To avoid `dynamic_cast`, you could add a dedicated type field to the base:

```
enum class CallbackType { Device, Session };
struct RegisteredCallbackToken {
    CallbackType callbackType;
};
struct DeviceRegisteredCallbackToken : public RegisteredCallbackToken {
    DeviceRegisteredCallbackToken : RegisteredCallbackToken({ callbackType = CallbackType.Device}) {}
    std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> Callbacks;
};
struct SessionRegisteredCallbackToken : public RegisteredCallbackToken {
    SessionRegisteredCallbackToken() : RegisteredCallbackToken({ callbackType = CallbackType.Session}) {}
    std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> Callbacks;
};
```
Though, this is not as robust.

Last, you could avoid polymorphism all-together by using a `std::variant` in `RegisteredCallbackToken` instead:

```
struct RegisteredCallbackToken {
    std::variant<
        std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks>,
        std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks>> Callbacks;
};
```

This would require `std::visit` to determine the actual type, but that then allows compile-time determination using `constexpr`. I think this approach is pretty clean.

Can you consider these points for reducing complexity and consolidation and let me know whether you think any of them are worth doing?

windows/devices/uwb/UwbConnector.cxx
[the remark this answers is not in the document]
forwardpointer: pushed an update that addresses the extra complexity

windows/devices/uwb/UwbConnector.cxx
forwardpointer: erase_if looks at all the elements so I don't think it's more efficient (although it does make for nice one-liner). I believe the way I implemented it will break the for loop immediately after erasing, so it never executes it++ on an invalid iterator.

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
forwardpointer: I didn't realize that the address of the token object could be used as a unique identifier, which is what the current design was built around. So, I'll make another PR that will address this fact, and I probably will use a lot of the suggestions here. Thanks!

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -663,20 +664,30 @@ UwbConnector::GetResolvedSessionEventCallbacks(uint32_t sessionId)
...
-    for (auto it = std::begin(sessionEventCallbacksWeak); it != std::end(sessionEventCallbacksWeak);) {
-        auto& sessionEventCallbackWeak = *it;
+    for (auto it = std::begin(sessionEventCallbackIds); it != std::end(sessionEventCallbackIds);) {
+        auto& sessionEventCallbackId = *it;
+        auto result = m_sessionEventCallbacksIdMap.find(sessionEventCallbackId);
+        if (result == std::cend(m_sessionEventCallbacksIdMap)) {
+            PLOG_ERROR << "could not find a sessionEventCallback id in the id map when it is present in m_sessionIdToCallbackIdsMap. id=" << sessionEventCallbackId;
+            return {};
+        }
+        auto& [callbackId, sessionIdAndsessionEventCallbackWeak] = *result;
```
corbin-phipps, line 677: nit: `sessionIdAndsessionEventCallbackWeak` -> `sessionIdAndSessionEventCallbackWeak`

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
forwardpointer: I just pushed some code that simplifies much of the code. I used the polymorphism approach, since I believe it makes the most sense given that we have separated session and device callbacks in the rest of the interface.

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -23,8 +23,23 @@ namespace uwb
 {
 struct RegisteredCallbackToken
 {
-    uint32_t callbackId;
-    bool isDeviceEventCallback;
+    virtual ~RegisteredCallbackToken(){};
```
abeltrano, line 26: nit: 
```suggestion
    virtual ~RegisteredCallbackToken() = default;
```

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -23,8 +23,23 @@ namespace uwb
...
-    bool isDeviceEventCallback;
+    virtual ~RegisteredCallbackToken(){};
+};
+
+struct RegisteredSessionCallbackToken : public RegisteredCallbackToken
+{
+    RegisteredSessionCallbackToken(uint32_t sessionId, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> Callbacks) :
+        sessionId(sessionId),
+        Callbacks(Callbacks){};
+    uint32_t sessionId;
```
abeltrano, line 34: style consistency:
```suggestion
    uint32_t SessionId;

```

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -913,99 +922,68 @@ UwbConnector::RegisterSessionEventCallbacks(uint32_t sessionId, std::weak_ptr<::
...
-        auto result = m_sessionEventCallbacksIdMap.find(callbackId);
-        if (result == std::cend(m_sessionEventCallbacksIdMap)) {
-            return;
-        }
-        m_sessionEventCallbacksIdMap.erase(result);
-
-        // now remove the id from the list of m_sessionIdToCallbackIdsMap
-        auto& [_, pair] = *result;
-        auto& [sessionId, callbackWeak] = pair;
+        auto sessionCallback = dynamic_pointer_cast<::uwb::RegisteredSessionCallbackToken>(tokenShared);
```
abeltrano, line 948: Right now `sessionCallback` will always be non-`nullptr` since we only have 2 derived types, but we should guard against the case that ever changes. So, let's add an additional check here:
```suggestion
        auto sessionCallback = dynamic_pointer_cast<::uwb::RegisteredSessionCallbackToken>(tokenShared);
        if (sessionCallback == nullptr) {
            throw std::runtime_error("invalid callback type used, this is a bug!");
        }
```

lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
```diff
@@ -17,6 +17,10 @@ namespace uwb
  */
 class RegisteredCallbackToken;
 
+class RegisteredSessionCallbackToken;
+
+class RegisteredDeviceCallbackToken;
+
```
abeltrano, line 23: Let's remove these since they're internal details and so the interface clients shouldn't be aware of them.
```suggestion
```

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -21,9 +21,25 @@ using namespace ::uwb::protocol::fira;
...
 {
-    uint32_t callbackId;
+    virtual ~RegisteredCallbackToken(){};
+};
+
+struct RegisteredSessionCallbackToken : public RegisteredCallbackToken
+{
+    RegisteredSessionCallbackToken(uint32_t sessionId, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> Callbacks) :
+        sessionId(sessionId),
+        Callbacks(Callbacks){};
```
abeltrano, line 33: nit: might as well move the callbacks since the expense of a copy has already been incurred. Also for clarity, the argument name should be lowercase. Same for `RegisteredDeviceCallbackToken` ctor.
```suggestion
    RegisteredSessionCallbackToken(uint32_t sessionId, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbacks> callbacks) :
        sessionId(sessionId),
        Callbacks(std::move(callbacks)){};
```

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -875,48 +895,95 @@ UwbConnector::NotificationListenerStop()
...
+UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackToken> token)
 {
-    // TODO implement
+    auto tokenShared = token.lock();
+    if (not tokenShared) {
+        return;
+    }
     std::lock_guard eventCallbacksLockExclusive{ m_eventCallbacksGate };
+
+    auto deviceCallback = std::dynamic_pointer_cast<::uwb::RegisteredDeviceCallbackToken>(tokenShared);
```
abeltrano, line 937: nit: It might be worth reversing the order of the checks to do the `::uwb::RegisteredSessionCallbackToken` type first since we would expect the callbacks for sessions to come and go regularly, but the ones for device to remain fairly static. Basically, put the likely branch condition first.

windows/devices/uwb/UwbConnector.cxx
abeltrano: I totally missed the `break`. So, you're right on both points: `erase_if` is less efficient and the way you did it was 100% safe.

👍

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
abeltrano: Wow, this was really well done, I love it!

abeltrano (approved):
Just a few small suggestions and nit picks, but this is completely awesome 💯

corbin-phipps (dismissed):
Still some minor changes that could be made (such as changing return type of InsertDeviceEventCallback to std::weak_ptr), but nothing that affects functionality, so it seems good to go

lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
forwardpointer: I pushed a change implementing this, but I'm wondering if it even makes sense to have one deregistration function, while there are two registration functions. I'll think about this more when I make a pr calling the deregistration function where it's needed.

abeltrano: Having 2 deregistration functions seems reasonable to me.

abeltrano (approved):

PR #226, opened 2023-04-13
Dereg individual

forwardpointer:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [X] Infrastructure

### Goals

Allow the deregistration of individual callbacks.

### Technical Details

Since some callback function signatures are ambiguous, I decided to make separate types for each callback. These types are used internally by UwbConnector to keep track of what's what.

On the client side (i.e. UwbDevice and UwbSession), things largely stayed the same. They still fill out the structure of device callbacks or session callbacks, which they pass into Register[Device|Session]EventCallbacks. However, now those functions return a structure containing weak_ptr to the callback tokens corresponding to each of the callbacks they registered. If the client leaves some callbacks out of the structure, then the callback token for those callbacks will be nullptr (and the UwbConnector won't have added that token to its internal storage). The structure containing all session or device callbacks is nice to have in the interface because it allows the client to unambiguously label the callbacks as what they are, without needing to be aware of all the internal types UwbConnector uses to keep track.

### Test Results
Regression tests pass

### Reviewer Focus

Are there any other approaches to this that avoid the problem of having to define a type for each callback?

### Future Work

* Actually call the deregistration in the TODO places
* Support recursive locks so that the callback can deregister itself.

### Checklist

### DIFF
--- a/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
+++ b/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
@@ -40,21 +40,21 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param reason The reason the session ended.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnSessionEnded> OnSessionEnded;
+    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionEnded> OnSessionEnded;
 
     /**
      * @brief Invoked when active ranging starts.
      *
      * @param session The session for which the event occurred.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnRangingStarted> OnRangingStarted;
+    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnRangingStarted> OnRangingStarted;
 
     /**
      * @brief Invoked when active ranging stops.
      *
      * @param session The session for which the event occurred.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnRangingStopped> OnRangingStopped;
+    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnRangingStopped> OnRangingStopped;
 
     /**
      * @brief Invoked when the properties of a peer involved in the session
@@ -63,7 +63,7 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param peersChanged A list of peers whose properties changed.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnPeerPropertiesChanged> OnPeerPropertiesChanged;
+    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnPeerPropertiesChanged> OnPeerPropertiesChanged;
 
     /**
      * @brief Invoked when membership of one or more near peers involved in
@@ -74,7 +74,7 @@ struct UwbRegisteredSessionEventCallbacks
      * @param peersAdded A list of peers that were added to the session.
      * @param peersRemoved A list of peers that were removed from the session.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnSessionMembershipChanged> OnSessionMembershipChanged;
+    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionMembershipChanged> OnSessionMembershipChanged;
 };
 
 /**
@@ -88,21 +88,21 @@ struct UwbRegisteredDeviceEventCallbacks
      *
      * @param status The generic error that occurred.
      */
-    std::shared_ptr<UwbRegisteredDeviceEventCallbacks::OnStatusChanged> OnStatusChanged;
+    std::shared_ptr<UwbRegisteredDeviceEventCallbackTypes::OnStatusChanged> OnStatusChanged;
 
     /**
      * @brief Invoked when the device status changes.
      *
      * @param statusDevice
      */
-    std::shared_ptr<UwbRegisteredDeviceEventCallbacks::OnDeviceStatusChanged> OnDeviceStatusChanged;
+    std::shared_ptr<UwbRegisteredDeviceEventCallbackTypes::OnDeviceStatusChanged> OnDeviceStatusChanged;
 
     /**
      * @brief Invoked when the status of a session changes.
      *
      * @param statusSession The new status of the session.
      */
-    std::shared_ptr<UwbRegisteredDeviceEventCallbacks::OnSessionStatusChanged> OnSessionStatusChanged;
+    std::shared_ptr<UwbRegisteredDeviceEventCallbackTypes::OnSessionStatusChanged> OnSessionStatusChanged;
 };
 
 /**
--- a/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
+++ b/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
@@ -11,6 +11,22 @@
 
 namespace uwb
 {
+namespace UwbRegisteredSessionEventCallbackTypes
+{
+using OnSessionEnded = std::function<void(::uwb::UwbSessionEndReason reason)>;
+using OnRangingStarted = std::function<void()>;
+using OnRangingStopped = std::function<void()>;
+using OnPeerPropertiesChanged = std::function<void(const std::vector<UwbPeer> peersChanged)>;
+using OnSessionMembershipChanged = std::function<void(const std::vector<UwbPeer> peersAdded, const std::vector<UwbPeer> peersRemoved)>;
+}; // namespace UwbRegisteredSessionEventCallbackTypes
+
+namespace UwbRegisteredDeviceEventCallbackTypes
+{
+using OnStatusChanged = std::function<void(::uwb::protocol::fira::UwbStatus)>;
+using OnDeviceStatusChanged = std::function<void(::uwb::protocol::fira::UwbStatusDevice)>;
+using OnSessionStatusChanged = std::function<void(::uwb::protocol::fira::UwbSessionStatus)>;
+}; // namespace UwbRegisteredDeviceEventCallbackTypes
+
 /**
  * @brief Interface for receiving events from a UwbSession. This is the primary
  * method to receive information from near peers.
@@ -24,21 +40,21 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param reason The reason the session ended.
      */
-    std::function<void(UwbSessionEndReason reason)> OnSessionEnded;
+    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnSessionEnded> OnSessionEnded;
 
     /**
      * @brief Invoked when active ranging starts.
      *
      * @param session The session for which the event occurred.
      */
-    std::function<void()> OnRangingStarted;
+    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnRangingStarted> OnRangingStarted;
 
     /**
      * @brief Invoked when active ranging stops.
      *
      * @param session The session for which the event occurred.
      */
-    std::function<void()> OnRangingStopped;
+    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnRangingStopped> OnRangingStopped;
 
     /**
      * @brief Invoked when the properties of a peer involved in the session
@@ -47,7 +63,7 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param peersChanged A list of peers whose properties changed.
      */
-    std::function<void(const std::vector<UwbPeer> peersChanged)> OnPeerPropertiesChanged;
+    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnPeerPropertiesChanged> OnPeerPropertiesChanged;
 
     /**
      * @brief Invoked when membership of one or more near peers involved in
@@ -58,7 +74,7 @@ struct UwbRegisteredSessionEventCallbacks
      * @param peersAdded A list of peers that were added to the session.
      * @param peersRemoved A list of peers that were removed from the session.
      */
-    std::function<void(const std::vector<UwbPeer> peersAdded, const std::vector<UwbPeer> peersRemoved)> OnSessionMembershipChanged;
+    std::shared_ptr<UwbRegisteredSessionEventCallbacks::OnSessionMembershipChanged> OnSessionMembershipChanged;
 };
 
 /**
@@ -72,21 +88,21 @@ struct UwbRegisteredDeviceEventCallbacks
      *
      * @param status The generic error that occurred.
      */
-    std::function<void(::uwb::protocol::fira::UwbStatus)> OnStatusChanged;
+    std::shared_ptr<UwbRegisteredDeviceEventCallbacks::OnStatusChanged> OnStatusChanged;
 
     /**
      * @brief Invoked when the device status changes.
      *
      * @param statusDevice
      */
-    std::function<void(::uwb::protocol::fira::UwbStatusDevice)> OnDeviceStatusChanged;
+    std::shared_ptr<UwbRegisteredDeviceEventCallbacks::OnDeviceStatusChanged> OnDeviceStatusChanged;
 
     /**
      * @brief Invoked when the status of a session changes.
      *
      * @param statusSession The new status of the session.
      */
-    std::function<void(::uwb::protocol::fira::UwbSessionStatus)> OnSessionStatusChanged;
+    std::shared_ptr<UwbRegisteredDeviceEventCallbacks::OnSessionStatusChanged> OnSessionStatusChanged;
 };
 
 /**
--- a/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
+++ b/lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
@@ -40,21 +40,21 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param reason The reason the session ended.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionEnded> OnSessionEnded;
+    std::weak_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionEnded> OnSessionEnded;
 
     /**
      * @brief Invoked when active ranging starts.
      *
      * @param session The session for which the event occurred.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnRangingStarted> OnRangingStarted;
+    std::weak_ptr<UwbRegisteredSessionEventCallbackTypes::OnRangingStarted> OnRangingStarted;
 
     /**
      * @brief Invoked when active ranging stops.
      *
      * @param session The session for which the event occurred.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnRangingStopped> OnRangingStopped;
+    std::weak_ptr<UwbRegisteredSessionEventCallbackTypes::OnRangingStopped> OnRangingStopped;
 
     /**
      * @brief Invoked when the properties of a peer involved in the session
@@ -63,7 +63,7 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param peersChanged A list of peers whose properties changed.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnPeerPropertiesChanged> OnPeerPropertiesChanged;
+    std::weak_ptr<UwbRegisteredSessionEventCallbackTypes::OnPeerPropertiesChanged> OnPeerPropertiesChanged;
 
     /**
      * @brief Invoked when membership of one or more near peers involved in
@@ -74,7 +74,7 @@ struct UwbRegisteredSessionEventCallbacks
      * @param peersAdded A list of peers that were added to the session.
      * @param peersRemoved A list of peers that were removed from the session.
      */
-    std::shared_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionMembershipChanged> OnSessionMembershipChanged;
+    std::weak_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionMembershipChanged> OnSessionMembershipChanged;
 };
 
 /**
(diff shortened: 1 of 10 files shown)


lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
```diff
@@ -30,21 +40,21 @@ struct UwbRegisteredSessionEventCallbacks
      * @param session The session for which the event occurred.
      * @param reason The reason the session ended.
```
corbin-phipps, line 41: This comment should be updated since OnSessionEnded is now a `std::weak_ptr`. Same with the other event callbacks

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -641,111 +692,90 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
...
-std::vector<std::shared_ptr<::uwb::UwbRegisteredDeviceEventCallbacks>>
-UwbConnector::GetResolvedDeviceEventCallbacks()
+/**
+ * @brief Internal helper function to invoke the callbacks with the arg
+ * If the weak_ptr callback held within a token is not valid, then the token is removed
+ *
+ * @tparam ArgT the type of the arg
+ * @tparam TokenT the type of the token
+ * @param tokens
+ * @param arg
```
corbin-phipps, line 702: nit: args

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -641,111 +692,90 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
     LOG_INFO << "uwb notification listener stopped for device " << DeviceName();
 }
 
-std::vector<std::shared_ptr<::uwb::UwbRegisteredDeviceEventCallbacks>>
-UwbConnector::GetResolvedDeviceEventCallbacks()
+/**
+ * @brief Internal helper function to invoke the callbacks with the arg
+ * If the weak_ptr callback held within a token is not valid, then the token is removed
+ *
+ * @tparam ArgT the type of the arg
```
corbin-phipps, line 699: nit: ArgTs

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+::uwb::UwbRegisteredDeviceEventCallbackTokens
+UwbConnector::RegisterDeviceEventCallbacks(::uwb::UwbRegisteredDeviceEventCallbacks callbacks)
 {
     std::lock_guard eventCallbacksLockExclusive{ m_eventCallbacksGate };
     bool isFirstCallback = not CallbacksPresent();
-    auto token = std::make_shared<::uwb::RegisteredSessionCallbackToken>(sessionId, callbacks);
-    auto node = m_sessionEventCallbacks.extract(sessionId);
+
+    auto OnStatusChangedToken = GetToken<::uwb::UwbRegisteredDeviceEventCallbackTypes::OnStatusChanged>(
+        callbacks, [](auto&& cbs) {
```
corbin-phipps, line 903: nit: Full word instead of cbs. Obviously can't use "callbacks", but maybe "eventCallbacks" or something similar

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+            return cbs.OnStatusChanged;
+        },
+        [this](auto&& cb) {
```
corbin-phipps, line 906: nit: cb -> callback

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+ * @brief Internal helper function to insert a token into the map of session tokens
+ *
+ * @tparam T the type of the token
+ * @param tokens the map of tokens to insert into
+ * @param sessionId
+ * @param token
+ */
+template <typename T>
+void
+InsertSessionToken(std::unordered_map<uint32_t, std::vector<std::shared_ptr<T>>>& tokens, uint32_t sessionId, std::shared_ptr<T> token)
```
corbin-phipps, line 949: Could sessionId could be const here?

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+ * @tparam L the type of the callback lambda
+ * @param sessionId
+ * @param callbacks the struct holding the callbacks
+ * @param accessor the accessor for the specific callback
+ * @param tokenizer the function that actually tokenizes the callback and gives ownership to the UwbConnector
+ * @return std::weak_ptr<::uwb::RegisteredSessionCallbackToken>
+ */
+template <typename L>
+std::weak_ptr<::uwb::RegisteredSessionCallbackToken>
+GetToken(uint32_t sessionId, ::uwb::UwbRegisteredSessionEventCallbacks callbacks, std::function<std::weak_ptr<L>(::uwb::UwbRegisteredSessionEventCallbacks)> accessor, std::function<std::weak_ptr<::uwb::RegisteredSessionCallbackToken>(uint32_t, std::weak_ptr<L>)> tokenizer)
```
corbin-phipps, line 973: Could sessionId be const here?

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+        OnPeerPropertiesChangedToken,
+        OnSessionMembershipChangedToken
+    };
 }
 
 bool
 UwbConnector::CallbacksPresent()
 {
-    return (not m_sessionEventCallbacks.empty()) or (not m_deviceEventCallbacks.empty());
+    return not(m_onSessionEndedCallbacks.empty() and m_onRangingStartedCallbacks.empty() and
```
corbin-phipps, line 1054: Could use `std::any_of`:

`return std::any_of(std::cbegin({a, b, c, d, e, f, g, h}), std::cend({a, b, c, d, e, f, g, h}), [](const auto& callback) { 
return !callback.empty(); 
});`

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+        m_onRangingStoppedCallbacks.empty() and m_onPeerPropertiesChangedCallbacks.empty() and m_onSessionMembershipChangedCallbacks.empty() and
+        m_onStatusChangedCallbacks.empty() and m_onDeviceStatusChangedCallbacks.empty() and m_onSessionStatusChangedCallbacks.empty());
+}
+
+/**
+ * @brief Internal helper function to try to deregister this token as this type
+ *
+ * @tparam T the type to try to dynamic_cast this token to
+ * @param token
+ * @param tokens
```
corbin-phipps, line 1064: Should be tokensMap

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -948,31 +1134,22 @@ UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackTok
...
-        auto tokens = node.mapped();
-        auto tokenIt = std::find_if(std::cbegin(tokens), std::cend(tokens), [sessionCallback](const auto& token) {
-            return token.get() == sessionCallback.get();
-        });
-        if (tokenIt == std::cend(tokens)) {
-            return; // no associated token found, bail
-        }
-        tokens.erase(tokenIt);
-        m_sessionEventCallbacks.insert(std::move(node));
+        DeregisterSessionEventCallback<::uwb::OnSessionEndedToken>(tokenShared, m_onSessionEndedCallbacks)                               ? 0
```
corbin-phipps, line 1137: I don't quite get the use of the ternary operators here. Can't you just do:

`DeregisterSessionEventCallback<::uwb::OnSessionEndedToken>(tokenShared, m_onSessionEndedCallbacks);`
`DeregisterSessionEventCallback<::uwb::OnRangingStartedToken>(tokenShared, m_onRangingStartedCallbacks);`
`DeregisterSessionEventCallback<::uwb::OnRangingStoppedToken>(tokenShared, m_onRangingStoppedCallbacks);`
...

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -948,31 +1134,22 @@ UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackTok
...
+            : DeregisterSessionEventCallback<::uwb::OnRangingStartedToken>(tokenShared, m_onRangingStartedCallbacks)                     ? 0
+            : DeregisterSessionEventCallback<::uwb::OnRangingStoppedToken>(tokenShared, m_onRangingStoppedCallbacks)                     ? 0
+            : DeregisterSessionEventCallback<::uwb::OnPeerPropertiesChangedToken>(tokenShared, m_onPeerPropertiesChangedCallbacks)       ? 0
+            : DeregisterSessionEventCallback<::uwb::OnSessionMembershipChangedToken>(tokenShared, m_onSessionMembershipChangedCallbacks) ? 0
+                                                                                                                                         : 0;
+        return;
```
corbin-phipps, line 1143: Is the `return` necessary here?

corbin-phipps (changes requested):

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -895,43 +868,256 @@ UwbConnector::NotificationListenerStop()
...
+}
+
+/**
+ * @brief Internal helper function that tokenizes a callback if it can be resolved
+ *
+ * @tparam L the type of the callback lambda
+ * @param sessionId
+ * @param callbacks the struct holding the callbacks
+ * @param accessor the accessor for the specific callback
+ * @param tokenizer the function that actually tokenizes the callback and gives ownership to the UwbConnector
```
abeltrano, line 968: Can we call this `createToken` or `createTokenFunc` instead (verb vs noun)?

windows/devices/uwb/UwbConnector.cxx
abeltrano: +1, not needed, same with other instance below.

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
```diff
@@ -68,24 +78,26 @@ public:
...
      * @brief Sets the callbacks for the UwbDevice that owns this UwbConnector
+     * You can pass in a partially filled in struct, in which case this function returns a partially filled in struct of tokens corresponding to those callbacks.
      *
      * @param callbacks
-     * @return std::weak_ptr<::uwb::RegisteredCallbackToken> You can pass this pointer into DeregisterEventCallback to deregister this event callback
+     * @return ::uwb::UwbRegisteredDeviceEventCallbackTokens You can pass the pointers into DeregisterEventCallback to deregister the event callbacks
      */
-    virtual std::weak_ptr<::uwb::RegisteredCallbackToken>
-    RegisterDeviceEventCallbacks(std::weak_ptr<::uwb::UwbRegisteredDeviceEventCallbacks> callbacks) override;
+    virtual ::uwb::UwbRegisteredDeviceEventCallbackTokens
```
abeltrano, line 86: Given we've now made all the callbacks individual, I wonder if we should expose individual functions to register them? We could keep this one for callers that want to aggregate, but it would be nice for callers that are only interested in 1 or 2 callbacks to just call individual functions.

abeltrano (changes requested):
This is really well done. I added some comments with suggestions for consolidating stuff and simplifying the callback instances themselves, but otherwise this looks great.

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: Makes sense. I was thinking about it conceptually - I wouldn't expect the sessionId to ever need to change in this function. But I totally understand having non-`const` in function definitions.

windows/devices/uwb/UwbConnector.cxx
forwardpointer: I don't think I can use any_of here because unfortunately the vectors have different types

windows/devices/uwb/UwbConnector.cxx
forwardpointer: The idea is to return early when it finds the correct class to cast to. I guess I could make it clearer by using if statements, but I think ternary statements are more concise.

lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
```diff
@@ -30,21 +40,21 @@ struct UwbRegisteredSessionEventCallbacks
...
      */
-    std::function<void(UwbSessionEndReason reason)> OnSessionEnded;
+    std::weak_ptr<UwbRegisteredSessionEventCallbackTypes::OnSessionEnded> OnSessionEnded;
```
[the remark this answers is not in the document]
forwardpointer: That's an interesting idea. I'll look into it

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: I actually think it might work still since they all have a `.empty()` method and are all iterable with `std::cbegin`. I'm definitely not positive so I could be wrong, but it might be worth a try. The current code is just a bit difficult to understand the logic and using `std::any_of` would make it clearer

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: Oh I see. Well in that case, I think you could just use `||` to short-circuit the code:

`DeregisterSessionEventCallback<::uwb::OnSessionEndedToken>(tokenShared, m_onSessionEndedCallbacks) || DeregisterSessionEventCallback<::uwb::OnRangingStartedToken>(tokenShared, m_onRangingStartedCallbacks) || ...`

windows/devices/uwb/UwbConnector.cxx
forwardpointer: I tried it and it wouldn't compile. I believe it doesn't compile because std::any_of uses an iterator over whatever container its checking. However, the type of the iterator is ill-defined since the vectors are distinct types. It _might_ be able to work if I construct new vectors which just cast the members of the original vectors to RegisteredCallbackToken, but I don't want to incur that additional cost. Basically the problem is that the tokens like OnSessionEndedToken are derived classes of RegisteredCallbackToken but std::vector<OnSessionEndedToken> is not a derived class of std::vector<RegisteredCallbackToken>

corbin-phipps: Ohh I see, well thanks for giving it a try!

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: Did this get updated?

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: Did this get updated?

windows/devices/uwb/UwbConnector.cxx
abeltrano: I think the types of each container being different is the limiting factor. `std::any_of` (and `std::ranges::any_of`) require a predicate which takes a certain type which is applied to each element of the input. I don't know how multiple vectors of different types could be adapted to take the same predicate.

@corbin-phipps There's certainly a way to do it with a traits-based approach as you described since all of the containers have an `empty()` function, but I don't see how it could be done with the STL constrained algorithms generically because you still need a way to group all of your inputs; the only way I know to do that is by combing the calls manually with `std::any_of(container<T>, pred<T>) && std::any_of(container<U>, pred<U>) && ...`.

You could do that combination elegantly with a template and C++ fold expressions, but I'm not sure it's worth the complexity. Something like (not sure if syntax is exactly correct here):

```
template <typename ContainerT, typename... Remaining>
requires <logic to ensure ContainerT has empty() function>
bool
AnyAreEmpty(const ContainerT& container, Remaining&&.. remainingContainers) {
    return (container.empty() || ...);
}
```

windows/devices/uwb/UwbConnector.cxx
forwardpointer: oops sorry, I didn't check if I updated it for all of them. I just pushed a change for it: let me know if I missed any.

corbin-phipps (approved):

lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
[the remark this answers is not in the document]
forwardpointer: Ultimately I decided to keep the current design. The main reason being that if we go with the weak_from_this approach, the client would have to do the same weak_from_this boilerplate code in every lambda they pass to the UwbConnector. This boilerplate code isn't immediately apparent when you look at the interface UwbRegisteredCallbacks.hxx, as it just says the callback is a naked std::function. However, I think it is a bit more apparent if the interface has weak_ptr to std::function, that the shared_ptr needs to be saved by the client. I do agree that either way the documentation should be updated to clearly reflect the usage.

lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
forwardpointer: updated docs

windows/devices/uwb/include/windows/devices/uwb/UwbConnector.hxx
forwardpointer: opened an issue to track this #231

corbin-phipps (dismissed):

forwardpointer:
I just ran a regression test on the basic ranging scenario, and it passes. So it should be ready for merging now

windows/devices/uwb/UwbConnector.cxx
forwardpointer: wow, I had no idea that C++ fold expressions existed. Great to know for the future in case we might need them!

lib/uwb/include/uwb/UwbRegisteredCallbacks.hxx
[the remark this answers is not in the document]
abeltrano: That's good reasoning. I agree it's more obvious the caller needs to manage the lifetime when the argument is a `weak_ptr`. So, I'm definitely good with keeping this as you have it.

For discussion purposes only, whenever one is using anything with a capture list (lambda, `std::function`, etc.), then they (should) automatically consider the lifetime of all the stuff being captured. So, I'd say that it's a reasonable expectation for callers to safely manage lifetime when using naked `std::function`.

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -641,111 +692,90 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
...
+        if (not callbackShared) {
+            it = tokens.erase(it);
         } else {
-            it = m_deviceEventCallbacks.erase(it);
+            auto callback = *callbackShared;
+            if (not callback) {
+                it = tokens.erase(it);
+            }
+            callback(args...);
+            it = std::next(it);
```
abeltrano, line 724: Should invocation of the callback and advancing the iterator be in an `else` clause? Otherwise, it seems `callback` is de-referenced when it's null and `it` is advanced twice (once with `erase`, another with `std::next(it)`). 

Also, I believe `args` should be forwarded to optimize out copies (this necessitates making `args` a universal reference, `ArgTs&&... args`):

```suggestion
            if (not callback) {
                it = tokens.erase(it);
            } else {
                callback(std::forward<ArgTs>(args));
                it = std::next(it);
            }
```

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -943,36 +1137,29 @@ UwbConnector::DeregisterEventCallback(std::weak_ptr<::uwb::RegisteredCallbackTok
...
 
-        auto node = m_sessionEventCallbacks.extract(sessionId);
-        if (node.empty()) {
-            return; // sessionId has no associated tokens
-        }
+        DeregisterSessionEventCallback<::uwb::OnSessionEndedToken>(tokenShared, m_onSessionEndedCallbacks) or
+            DeregisterSessionEventCallback<::uwb::OnRangingStartedToken>(tokenShared, m_onRangingStartedCallbacks) or
+            DeregisterSessionEventCallback<::uwb::OnRangingStoppedToken>(tokenShared, m_onRangingStoppedCallbacks) or
+            DeregisterSessionEventCallback<::uwb::OnPeerPropertiesChangedToken>(tokenShared, m_onPeerPropertiesChangedCallbacks) or
+            DeregisterSessionEventCallback<::uwb::OnSessionMembershipChangedToken>(tokenShared, m_onSessionMembershipChangedCallbacks);
```
abeltrano, line 1151: It'd be nice if we could avoid this pattern. The type information is built in to the instance (`tokenShared`), so we could make use of the polymorphic nature of the token and have it de-register itself. For example, you could add a pure `virtual` method to `RegisteredSessionCallbackToken` like:

```
struct RegisteredSessionCallbackToken {
    virtual void Deregister() = 0;
```

Then have sub-classes implement that. So, something like this:
```
struct OnSessionEndedToken : public RegisteredSessionCallbackToken
{
    OnSessionEndedToken(uint32_t sessionId, std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbackTypes::OnSessionEnded> callback, std::unordered_map<uint32_t, std::vector<std::shared_ptr<::uwb::OnSessionEndedToken>>> &tokenMap) :
        RegisteredSessionCallbackToken(sessionId),
        Callback(std::move(callback)){};
    std::weak_ptr<::uwb::UwbRegisteredSessionEventCallbackTypes::OnSessionEnded> Callback;
    std::unordered_map<uint32_t, std::vector<std::shared_ptr<::uwb::OnSessionEndedToken>>>& TokenMap;
    
    void Deregister() override { DeregisterSessionEventCallback(this, TokenMap);
};
```

This would require updating `DeregisterSessionEventCallback` to take a `::uwb::RegisteredCallbackToken*` instead of the `shared_ptr<::uwb::RegisteredCallbackToken>`, but that's no issue at all.

I thought it would also be nice to have it encode the ability to execute the callback as well using the same pattern. Anyway, perhaps something to consider for a follow-up PR.

abeltrano (changes requested):

windows/devices/uwb/UwbConnector.cxx
forwardpointer: good suggestion, updated

abeltrano (approved):

windows/devices/uwb/UwbConnector.cxx
forwardpointer: resolving this to be addressed in the next pr

PR #239, opened 2023-04-19
small fix to check for a valid deregister function before executing it

forwardpointer:
### Type

- [X] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Checklist

### DIFF
--- a/windows/devices/uwb/UwbConnector.cxx
+++ b/windows/devices/uwb/UwbConnector.cxx
@@ -31,6 +31,10 @@ struct RegisteredCallbackToken
      */
     RegisteredCallbackToken(std::function<void(RegisteredCallbackToken*)> deregister) :
         Deregister([this, deregister = std::move(deregister)]() {
+            if (not deregister) {
+                PLOG_WARNING << "empty callback token";
+                return;
+            }
             return deregister(this);
         })
     {}


windows/devices/uwb/UwbConnector.cxx
```diff
@@ -31,6 +31,10 @@ struct RegisteredCallbackToken
      */
     RegisteredCallbackToken(std::function<void(RegisteredCallbackToken*)> deregister) :
         Deregister([this, deregister = std::move(deregister)]() {
+            if (not deregister) {
+                PLOG_WARNING << "empty callback token";
+                return;
+            }
```
abeltrano, line 37: When can/should this ever be the case?

forwardpointer: It shouldn't ever be the case but nothing prevents it from happening so I thought it would be better to log something if it is the case

forwardpointer: for example, if the deregister function was actually initialized with the default constructor std::function<>();

abeltrano (approved):

corbin-phipps (approved):

PR #242, opened 2023-04-20
Begin improving nocli service start command

corbin-phipps:
### Type

- [x] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

To prepare for utilizing the nocli tool to call into NearObject Service APIs, the "nocli.exe service range start" command needs to be cleaned up and improved, similar to the "nocli.exe uwb range start" command that was improved previously.

This PR focuses on the CLI options and their descriptions. Future PR(s) will ensure that the OOB parameters set through CLI options can all be translated to UCI ApplicationConfigurationParameters and displayed properly, as well as ensuring that a ranging session can run as expected.

### Technical Details

- Removed type labels (e.g. UINT) that were automatically added by CLI11.
- Added remaining session data parameters from the FiRa CSML spec.
- Updated the CLI option format and descriptions.
- Set MAC addresses based on current information.
- Improved printing out selected parameters.
- Updated MAGIC_ENUM_RANGE_MAX to 256 so that the UwbConfiguration::ParameterTag enum values can be displayed.
- Removed unused HostIsController boolean field, since that information is available in DeviceType.
- Fixed SlotDuration type to uint16_t (previously uint8_t).

### Test Results

Verified output in nocli.

### Reviewer Focus

None.

### Future Work

- Currently, the selected parameters that are displayed only include the ones that are part of UwbConfiguration (aka have a ParameterTag). There are other parameters (e.g. NumberOfControlees) that don't have a ParameterTag, but should be displayed. This needs to be solved.
- We still do not have a solution/understanding of the ControleeShortMacAddress situation, so we currently cannot utilize extended controlee mac addresses. This can be updated once we receive more information.
- There is likely more work to do in order to use OOB parameters from the CSML spec to configure a working ranging session, due to the differences between CSML-specified and UCI-specified parameters. I have not dug into this yet, so I'm unsure of the scope at the moment.
- The --help output shows "--StaticRangingInfoVendorId [0]". The "[0]" part needs to be removed, I'm not yet sure what causes it.
- Input validation of non-enum parameters needs to be added, similar to the UCI ApplicationConfigurationParameters that were validated for "nocli.exe uwb range start".

### Checklist


tools/cli/NearObjectCli.cxx
```diff
@@ -9,6 +9,9 @@
 #include <variant>
 #include <vector>
 
+#define MAGIC_ENUM_RANGE_MIN 0
+#define MAGIC_ENUM_RANGE_MAX 256
```
abeltrano, line 13: This will increase build times since it applies to everything in this file, not just `UwbConfiguration::ParameterTag` where it's needed. You can instead specialize `enum_range<UwbConfiguration::ParameterTag>` in the `customize` namespace.

We should also tighten the maximum range to reduce compile times as much as possible. It looks like the min is `FiraPhyVersion == 0x80` and max is `MaxRrRetry == 0x9D` is the max value for now, so can we use that?

```cpp
template <>
struct magic_enum::customize::enum_range<::uwb::protocol::fira::UwbConfiguration::ParameterTag> {
  static constexpr int min = 0x80; // Current minimum value is UwbConfiguration::ParameterTag::FiraPhyVersion == 0x80
  static constexpr int max = 0x9D; // Current maximum value is UwbConfiguration::ParameterTag::MaxRrRetry == 0x9D
};
```

abeltrano (changes requested):
Just one thing for perf, but otherwise this looks great.

forwardpointer (dismissed):

tools/cli/NearObjectCli.cxx
corbin-phipps: Oh good point, that would be better. I'll make that change.

corbin-phipps: I ended up adding this in NearObjectCli.hxx since NearObjectCli is where we use this. It may make sense to put this in UwbConfiguration instead, but I had issues with including `#include magic_enum.hpp`

abeltrano: It's ok here for now 👍

abeltrano (approved):

PR #243, opened 2023-04-21
Add peer app config params

forwardpointer:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Implement the command to add a peer before the ranging session is active.

### Technical Details

Added two new functions to the top-level UwbSession interface, which get and set a vector of app config params

### Test Results

Compile tested.

### Checklist

### DIFF
--- a/lib/uwb/UwbSession.cxx
+++ b/lib/uwb/UwbSession.cxx
@@ -137,6 +137,20 @@ UwbSession::GetApplicationConfigurationParameters()
     return GetApplicationConfigurationParametersImpl();
 }
 
+std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+UwbSession::GetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes)
+{
+    PLOG_VERBOSE << "get application configuration parameters";
+    return GetApplicationConfigurationParametersImpl(requestedTypes);
+}
+
+void
+UwbSession::SetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params)
+{
+    PLOG_VERBOSE << "get application configuration parameters";
+    return SetApplicationConfigurationParametersImpl(params);
+}
+
 UwbSessionState
 UwbSession::GetSessionState()
 {
--- a/lib/uwb/include/uwb/UwbSession.hxx
+++ b/lib/uwb/include/uwb/UwbSession.hxx
@@ -106,7 +106,7 @@ public:
 
     /**
      * @brief Attempt to add a controlee to this session.
-     * 
+     *
      * @param controleeMacAddress The mac address of the controlee. This is
      * expected to be in the mac address format configured for the session.
      * @return UwbStatus The status of the operation. UwbStatusGeneric::Ok is
@@ -142,6 +142,24 @@ public:
      */
     std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
     GetApplicationConfigurationParameters();
+
+    /**
+     * @brief Get the Application Configuration Parameters object
+     *
+     * @param requestedTypes
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+    GetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes);
+
+    /**
+     * @brief Get the Application Configuration Parameters object
+     *
+     * @param requestedTypes
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    void
+    SetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params);
 
     /**
      * @brief Get the current state for this session.
@@ -222,7 +240,7 @@ private:
 
     /**
      * @brief Attempt to add a controlee to this session.
-     * 
+     *
      * @param controleeMacAddress The mac address of the controlee. This is
      * expected to be in the mac address format configured for the session.
      * @return UwbStatus The status of the operation. UwbStatusGeneric::Ok is
@@ -238,6 +256,24 @@ private:
      */
     virtual std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
     GetApplicationConfigurationParametersImpl() = 0;
+
+    /**
+     * @brief Get the Application Configuration Parameters object
+     *
+     * @param requestedTypes
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    virtual std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) = 0;
+
+    /**
+     * @brief Get the Application Configuration Parameters object
+     *
+     * @param requestedTypes
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    virtual void
+    SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params) = 0;
 
     /**
      * @brief Get the current state for this session.
--- a/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
@@ -357,6 +357,16 @@ using UwbStatus = std::variant<UwbStatusGeneric, UwbStatusSession, UwbStatusRang
 bool
 IsUwbStatusOk(const UwbStatus& uwbStatus) noexcept;
 
+/**
+ * @brief Determines if the specified UWB status describes a retry command
+ *
+ * @param uwbStatus The status to check.
+ * @return true
+ * @return false
+ */
+bool
+IsUwbStatusRetry(const UwbStatus& uwbStatus) noexcept;
+
 enum class UwbStatusMulticast {
     OkUpdate,
     ErrorListFull,
--- a/windows/devices/uwb/UwbSession.cxx
+++ b/windows/devices/uwb/UwbSession.cxx
@@ -187,36 +187,29 @@ UwbSession::TryAddControleeImpl([[maybe_unused]] ::uwb::UwbMacAddress controleeM
 
     auto params = GetApplicationConfigurationParameters({ UwbApplicationConfigurationParameterType::DestinationMacAddresses });
 
-    // if (IsUwbStatusOk(uwbStatus)) {
-    {
-        std::unordered_set<::uwb::UwbMacAddress> macAddresses;
-        for (auto &param : params) {
-            auto pvalue = std::get_if<std::unordered_set<::uwb::UwbMacAddress>>(&(param.Value));
-            if (pvalue) {
-                macAddresses = *pvalue;
-                break;
-            }
+    std::unordered_set<::uwb::UwbMacAddress> macAddresses;
+    for (auto &param : params) {
+        auto pvalue = std::get_if<std::unordered_set<::uwb::UwbMacAddress>>(&(param.Value));
+        if (pvalue) {
+            macAddresses = *pvalue;
+            break;
         }
-        auto [_, inserted] = macAddresses.insert(controleeMacAddress);
-        if (not inserted) {
-            PLOG_INFO << "controleeMacAddress already added, skipping";
-            return UwbStatusGeneric::Ok;
-        }
-        if (macAddresses.size() > 8) {
-            PLOG_WARNING << "exceeded max number of controlees with numberOfControlees=" << macAddresses.size();
-            return UwbStatusGeneric::Rejected;
-        }
-        UwbApplicationConfigurationParameter dstMacAddresses{ .Type = UwbApplicationConfigurationParameterType::DestinationMacAddresses, .Value = macAddresses };
-        UwbApplicationConfigurationParameter numControlees{ .Type = UwbApplicationConfigurationParameterType::NumberOfControlees, .Value = uint8_t(macAddresses.size()) };
-        UwbStatus overallStatus = UwbStatusGeneric::CommandRetry;
-        while (IsUwbStatusRetry(overallStatus)) {
-            auto fut = m_uwbSessionConnector->SetApplicationConfigurationParameters(m_sessionId, { numControlees, dstMacAddresses });
-            auto [setStatus, paramStatuses] = fut.get();
-
-            overallStatus = setStatus;
-        }
-        return overallStatus;
     }
+    auto [_, inserted] = macAddresses.insert(controleeMacAddress);
+    if (not inserted) {
+        PLOG_INFO << "controleeMacAddress already added, skipping";
+        return UwbStatusGeneric::Ok;
+    }
+    if (macAddresses.size() > 8) {
+        PLOG_WARNING << "exceeded max number of controlees with numberOfControlees=" << macAddresses.size();
+        return UwbStatusGeneric::Rejected;
+    }
+    UwbApplicationConfigurationParameter dstMacAddresses{ .Type = UwbApplicationConfigurationParameterType::DestinationMacAddresses, .Value = macAddresses };
+    UwbApplicationConfigurationParameter numControlees{ .Type = UwbApplicationConfigurationParameterType::NumberOfControlees, .Value = uint8_t(macAddresses.size()) };
+
+    SetApplicationConfigurationParameters({ numControlees, dstMacAddresses });
+
+    return UwbStatusGeneric::Ok;
 }
 
 std::vector<UwbApplicationConfigurationParameter>
@@ -275,6 +268,27 @@ UwbSession::GetApplicationConfigurationParametersImpl(std::vector<::uwb::protoco
 void
 UwbSession::SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params)
 {
+    uint32_t sessionId = GetId();
+    while (true) {
+        auto resultFuture = m_uwbSessionConnector->SetApplicationConfigurationParameters(sessionId, params);
+        try {
+            auto [uwbStatus, applicationConfigurationParametersStatus] = resultFuture.get();
+
+            if (IsUwbStatusOk(uwbStatus)) {
+                return;
+            } else if (IsUwbStatusRetry(uwbStatus)) {
+                continue;
+            } else {
+                throw UwbException(uwbStatus);
+            }
+        } catch (UwbException &uwbException) {
+            PLOG_ERROR << "caught exception attempting to obtain application configuration parameters for session id " << sessionId << " (" << ToString(uwbException.Status) << ")";
+            throw uwbException;
+        } catch (std::exception &e) {
+            PLOG_ERROR << "caught unexpected exception attempting to obtain application configuration parameters for session id " << sessionId << " (" << e.what() << ")";
+            throw e;
+        }
+    }
 }
 
 UwbSessionState
--- a/windows/devices/uwb/UwbSession.cxx
+++ b/windows/devices/uwb/UwbSession.cxx
@@ -185,7 +185,7 @@ UwbSession::TryAddControleeImpl([[maybe_unused]] ::uwb::UwbMacAddress controleeM
         return UwbStatusGeneric::Rejected;
     }
 
-    auto params = GetApplicationConfigurationParameters();
+    auto params = GetApplicationConfigurationParameters({ UwbApplicationConfigurationParameterType::DestinationMacAddresses });
 
     // if (IsUwbStatusOk(uwbStatus)) {
     {
@@ -244,6 +244,37 @@ UwbSession::GetApplicationConfigurationParametersImpl()
         PLOG_ERROR << "caught unexpected exception attempting to obtain application configuration parameters for session id " << sessionId << " (" << e.what() << ")";
         throw e;
     }
+}
+
+std::vector<UwbApplicationConfigurationParameter>
+UwbSession::GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes)
+{
+    uint32_t sessionId = GetId();
+
+    auto resultFuture = m_uwbSessionConnector->GetApplicationConfigurationParameters(sessionId, requestedTypes);
+    if (!resultFuture.valid()) {
+        PLOG_ERROR << "failed to obtain application configuration parameters for session id " << sessionId;
+        throw UwbException(UwbStatusGeneric::Failed);
+    }
+
+    try {
+        auto [uwbStatus, applicationConfigurationParameters] = resultFuture.get();
+        if (!IsUwbStatusOk(uwbStatus)) {
+            throw UwbException(uwbStatus);
+        }
+        return applicationConfigurationParameters;
+    } catch (UwbException &uwbException) {
+        PLOG_ERROR << "caught exception attempting to obtain application configuration parameters for session id " << sessionId << " (" << ToString(uwbException.Status) << ")";
+        throw uwbException;
+    } catch (std::exception &e) {
+        PLOG_ERROR << "caught unexpected exception attempting to obtain application configuration parameters for session id " << sessionId << " (" << e.what() << ")";
+        throw e;
+    }
+}
+
+void
+UwbSession::SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params)
+{
 }
 
 UwbSessionState
--- a/windows/devices/uwb/UwbSession.cxx
+++ b/windows/devices/uwb/UwbSession.cxx
@@ -180,58 +180,43 @@ UwbSession::TryAddControleeImpl([[maybe_unused]] ::uwb::UwbMacAddress controleeM
     PLOG_VERBOSE << "TryAddControleeImpl";
     return UwbStatusGeneric::Rejected;
 
-    // TODO: convert code below to invoke IOCTL_UWB_SET_APP_CONFIG_PARAMS to use connector
+    if (not m_uwbSessionConnector) {
+        PLOG_WARNING << "No associated connector";
+        return UwbStatusGeneric::Rejected;
+    }
 
-    // TODO: two main options for updating the UWB-CLX peer list:
-    //  1) Every time a peer is added (on-demand)
-    //  2) Only when StartRanging() is called.
-    // Below sample code exemplifies 1) for simplicity but this is not necessarily the way to go.
-    //
-    // TODO: request UWB-CLX to update controlee list per below pseudo-code,
-    // which is *very* rough and some parts probably plain wrong:
-    //
+    auto params = GetApplicationConfigurationParameters();
 
-    // const auto macAddressLength = m_uwbMacAddressSelf.GetLength();
-    // const auto macAddressessLength = macAddressLength * m_peers.size();
+    // if (IsUwbStatusOk(uwbStatus)) {
+    {
+        std::unordered_set<::uwb::UwbMacAddress> macAddresses;
+        for (auto &param : params) {
+            auto pvalue = std::get_if<std::unordered_set<::uwb::UwbMacAddress>>(&(param.Value));
+            if (pvalue) {
+                macAddresses = *pvalue;
+                break;
+            }
+        }
+        auto [_, inserted] = macAddresses.insert(controleeMacAddress);
+        if (not inserted) {
+            PLOG_INFO << "controleeMacAddress already added, skipping";
+            return UwbStatusGeneric::Ok;
+        }
+        if (macAddresses.size() > 8) {
+            PLOG_WARNING << "exceeded max number of controlees with numberOfControlees=" << macAddresses.size();
+            return UwbStatusGeneric::Rejected;
+        }
+        UwbApplicationConfigurationParameter dstMacAddresses{ .Type = UwbApplicationConfigurationParameterType::DestinationMacAddresses, .Value = macAddresses };
+        UwbApplicationConfigurationParameter numControlees{ .Type = UwbApplicationConfigurationParameterType::NumberOfControlees, .Value = uint8_t(macAddresses.size()) };
+        UwbStatus overallStatus = UwbStatusGeneric::CommandRetry;
+        while (IsUwbStatusRetry(overallStatus)) {
+            auto fut = m_uwbSessionConnector->SetApplicationConfigurationParameters(m_sessionId, { numControlees, dstMacAddresses });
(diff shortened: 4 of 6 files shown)


windows/devices/uwb/UwbSession.cxx
```diff
@@ -180,58 +180,36 @@ UwbSession::TryAddControleeImpl([[maybe_unused]] ::uwb::UwbMacAddress controleeM
...
+            macAddresses = *pvalue;
+            break;
+        }
+    }
+    auto [_, inserted] = macAddresses.insert(controleeMacAddress);
+    if (not inserted) {
+        PLOG_INFO << "controleeMacAddress already added, skipping";
+        return UwbStatusGeneric::Ok;
+    }
+    if (macAddresses.size() > 8) {
```
forwardpointer, line 203: this constant should probably be in FiraDevice.hxx

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
```diff
@@ -602,6 +612,8 @@ constexpr std::array<UwbApplicationConfigurationParameterType, 5> UwbApplication
     UwbApplicationConfigurationParameterType::RangeDataNotificationProximityFar,
 };
 
+constexpr auto MAX_CONTROLEES = 8;
```
corbin-phipps, line 615: There is a place somewhere in one of the specs where it says the max number of controlees could be greater than 8 (@abeltrano do you remember where?). This came up in one of my previous PRs, otherwise this constant would already exist below line 60 (below `constexpr uint8_t MinimumNumberOfControlees = 1;`)

windows/devices/uwb/UwbSession.cxx
```diff
@@ -261,6 +239,58 @@ UwbSession::GetApplicationConfigurationParametersImpl()
...
+        PLOG_ERROR << "caught unexpected exception attempting to obtain application configuration parameters for session id " << sessionId << " (" << e.what() << ")";
+        throw e;
+    }
+}
+
+void
+UwbSession::SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params)
+{
+    uint32_t sessionId = GetId();
+    while (true) {
```
corbin-phipps, line 272: Why is this loop here? It looks like the idea is to keep trying to set the parameters if `IsUwbStatusRetry` is true, but when would that ever be the case? From my understanding, when we send the SetApplicationConfigurationParameters command to the device, it should return either ok, invalid parameter, invalid range, or error session active.

windows/devices/uwb/include/windows/devices/uwb/UwbSession.hxx
```diff
@@ -85,6 +85,24 @@ private:
...
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    virtual std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) override;
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
+     * @return void
```
corbin-phipps, line 101: I don't think the `@return void` line is necessary, based on other existing function comments in this file

windows/devices/uwb/UwbSession.cxx
```diff
@@ -180,58 +180,36 @@ UwbSession::TryAddControleeImpl([[maybe_unused]] ::uwb::UwbMacAddress controleeM
...
+    if (not m_uwbSessionConnector) {
+        PLOG_WARNING << "No associated connector";
+        return UwbStatusGeneric::Rejected;
+    }
+
+    auto params = GetApplicationConfigurationParameters({ UwbApplicationConfigurationParameterType::DestinationMacAddresses });
+
+    std::unordered_set<::uwb::UwbMacAddress> macAddresses;
+    for (auto &param : params) {
+        auto pvalue = std::get_if<std::unordered_set<::uwb::UwbMacAddress>>(&(param.Value));
```
corbin-phipps, line 192: What does `pvalue` mean? Is the `p` supposed to be for "pointer"? If so, we may want to avoid that kind of naming scheme since we don't use that anywhere else in the code base (to my knowledge, I could be wrong). If the `p` stands for "param", then let's change this to `paramValue`

corbin-phipps, line 192: I wonder if we could use something from `magic_enum` here instead

windows/devices/uwb/UwbSession.cxx
```diff
@@ -180,58 +180,36 @@ UwbSession::TryAddControleeImpl([[maybe_unused]] ::uwb::UwbMacAddress controleeM
...
+    if (not inserted) {
+        PLOG_INFO << "controleeMacAddress already added, skipping";
+        return UwbStatusGeneric::Ok;
+    }
+    if (macAddresses.size() > MAX_CONTROLEES) {
+        PLOG_WARNING << "exceeded max number of controlees with numberOfControlees=" << macAddresses.size();
+        return UwbStatusGeneric::Rejected;
+    }
+    UwbApplicationConfigurationParameter dstMacAddresses{ .Type = UwbApplicationConfigurationParameterType::DestinationMacAddresses, .Value = macAddresses };
+    UwbApplicationConfigurationParameter numControlees{ .Type = UwbApplicationConfigurationParameterType::NumberOfControlees, .Value = uint8_t(macAddresses.size()) };
```
corbin-phipps, line 208: Should use `static_cast<uint8_t>(macAddresses.size())` instead of the C-style cast

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -143,6 +143,24 @@ public:
...
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+    GetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes);
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
+     * @return void
```
corbin-phipps, line 159: I don't think the `@return void` line is necessary here, nor in the other locations

corbin-phipps (changes requested):

windows/devices/uwb/include/windows/devices/uwb/UwbSession.hxx
forwardpointer: looks like this was added by the docgen. i'll remove

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
abeltrano: That's right, it's in CSML Table 7, UWB Configuration Parameters from Controller:

image

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -239,6 +257,24 @@ private:
...
     GetApplicationConfigurationParametersImpl() = 0;
 
+    /**
+     * @brief Get the Application Configuration Parameters object
+     *
+     * @param requestedTypes
+     * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+     */
+    virtual std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) = 0;
```
abeltrano, line 267: This seems to already exist on line 258. Can we remove one?

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
abeltrano: Also, if we do need some kind of constant, please name it per style guidelines using PascalCase. Since this value is a little weird, adding some clarifying comments to the declaration/definition will definitely help:

```suggestion
/**
 * The minimum number of maximum controlees a particular implementation must handle. The maximum number of controlees may be higher if the implementation can support it.
 */
constexpr auto MaxControleesMin = 8;
```

(I'm not sure if the variable name I used above is the best, but you get the idea)

windows/devices/uwb/UwbSession.cxx
abeltrano: +1, but please use complete words as well:

```suggestion
        auto parameterValue = std::get_if<std::unordered_set<::uwb::UwbMacAddress>>(&(param.Value));
```

windows/devices/uwb/UwbSession.cxx
forwardpointer: After looking at the UCI spec again, I agree with what you're saying here. That being said, we don't currently handle CORE_GENERIC_ERROR_NTF from the UWBS, especially not when the Status is set to STATUS_COMMAND_RETRY. Opening #246 to track it

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
forwardpointer: Implementation can support more than 8 anchors, but the UCI spec doesn't mention this. Should we just follow the UCI spec since the command we're implementing is described there?

windows/devices/uwb/UwbSession.cxx
abeltrano, line 192: Since we only request 1 parameter, we should get either 0 or 1 back and so neither the loop nor the `get_if` should be needed. If we got back a mis-matched type or more than 1 result, it'd be a programmer error and we should just let it crash and burn. So, you should be able to do this instead:

```cpp
if (!params.empty()) {
    auto parameterValue = std::get<std::unordered_set<::uwb::UwbMacAddress>>(params.front().Value));
    // insert value into macAddresses ..
}
```

windows/devices/uwb/UwbSession.cxx
abeltrano: Also `std::size`
```suggestion
    UwbApplicationConfigurationParameter numControlees{ .Type = UwbApplicationConfigurationParameterType::NumberOfControlees, .Value = static_cast<uint8_t>(std::size(macAddresses)) };
```

windows/devices/uwb/UwbSession.cxx
abeltrano: Even though the spec explicitly calls out the status values for `SESSION_SET_APP_CONFIG_RSP` and retry is _not_ included, I think `STATUS_COMMAND_RETRY` can apply to any command. I'd rather handle it than not.

However, I'd also say that we should handle all such status indications the same way, and have configurable settings controlling it such as maximum number of retries, delay between retries, etc. I'd also rather see that done in a separate PR in the future when we're more focused on reliability. Let's just fail if the status is retry for now. Created #247 to track that work.

windows/devices/uwb/UwbSession.cxx
forwardpointer: We could case on the enum param.Type but we would still also have to dereference the variant param.Value as the actual type. That's why I just directly check the type of param.Value through std::get_if

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
corbin-phipps: I also recommend moving this constant to where the other similar constants are (around line 60 in my previous comment)

windows/devices/uwb/UwbSession.cxx
corbin-phipps: I see, since `STATUS_COMMAND_RETRY` is technically a possible value, then it should be handled. I just don't know when that would ever be the case, since it is never mentioned anywhere in the UCI spec (besides the list of status codes).

I definitely agree with having a max number of retries, since an infinite loop could be very bad. Even if it's likely that we're asked to retry a command an infinite number of times, it's still good to code defensively

windows/devices/uwb/UwbSession.cxx
corbin-phipps: I've said "param" so many times in my head, as well as out loud, that I almost thought it was an entire word and not short for "parameter" 😆

corbin-phipps: Oh right, I forgot that that's necessary for the `std::variant` value field. I think this is fine then

lib/uwb/include/uwb/UwbSession.hxx
forwardpointer: updated

windows/devices/uwb/UwbSession.cxx
forwardpointer: I think we should leave this to a future PR. For now, I'll have the function return an error when it requests retry through the status, as it's consistent with how the other functions handle it

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
abeltrano: > Implementation can support more than 8 anchors, but the UCI spec doesn't mention this. Should we just follow the UCI spec since the command we're implementing is described there?

I would probably just remove the code that's enforcing this maximum (and thus, remove this constant too). We should just let the driver handle it, which will be forwarded to the device which has the impl and it will indicate whether it's valid or not.

windows/devices/uwb/UwbSession.cxx
abeltrano: > I see. Because `STATUS_COMMAND_RETRY` is technically a possible value, then it should be handled. I just don't know when that would ever be the case, since it is never mentioned anywhere in the UCI spec (besides the list of status codes).

Given the spec is often ambiguous, my guess is that they didn't explicitly call out all the places `STATUS_COMMAND_RETRY` could be returned, and that it could be returned for any command. From my experience, it's not uncommon for simple hardware to return a "retry" status indication when it just can't handle the request at that specific moment, instead of implementing additional complex logic in the hardware/firmware to handle queuing/control flow of concurrent operations when they can only support sequential (this happens a lot with microcontrollers for cost-savings).

> I definitely agree with having a max number of retries, since an infinite loop could be very bad. Even if it's likely that we're asked to retry a command an infinite number of times, it's still good to code defensively

💯

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -232,12 +242,21 @@ private:
...
-    GetApplicationConfigurationParametersImpl() = 0;
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) = 0;
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
+     */
+    virtual void
+    SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params) = 0;
```
abeltrano, line 259: nit: Let's name this with the same pattern as others across the codebase:
```suggestion
    SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> uwbApplicationConfigurationParameters) = 0;
```

lib/uwb/include/uwb/UwbSession.hxx
corbin-phipps: This one is still here

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
```diff
@@ -75,6 +75,8 @@ constexpr uint8_t MaximumNumberOfStsSegmentsBprf = 1;
 constexpr uint8_t MaximumNumberOfStsSegmentsHprf = 4;
 constexpr uint8_t MinimumInBandTerminationAttemptCount = 0;
 constexpr uint8_t MaximumInBandTerminationAttemptCount = 10;
+constexpr uint8_t MinumumMandatoryMaximumNumberOfControleesInMulticastSession = 8;
```
corbin-phipps, line 78: Typo: Should be `MinimumMandatoryMaximumNumberOfControleesInMulticastSession`

windows/devices/uwb/UwbSession.cxx
```diff
@@ -178,69 +178,40 @@ UwbStatus
...
-    // // Attempt to set all new parameters.
-    // DWORD bytesReturned = 0;
-    // UWB_SET_APP_CONFIG_PARAMS_STATUS appConfigParamsStatus; // TODO: this needs to be dynamically allocated to fit returned content
-    // BOOL ioResult = DeviceIoControl(m_handleDriver.get(), IOCTL_UWB_SET_APP_CONFIG_PARAMS, &appConfigParams, static_cast<DWORD>(appConfigParamsSize), &appConfigParamsStatus, sizeof appConfigParamsStatus, &bytesReturned, nullptr);
-    // if (!LOG_IF_WIN32_BOOL_FALSE(ioResult)) {
-    //     // TODO
-    //     HRESULT hr = GetLastError();
-    //     PLOG_ERROR << "could not send params to driver, hr=" << std::showbase << std::hex << hr;
-    // }
+    if (not m_uwbSessionConnector) {
```
abeltrano, line 181: Let's be consistent with the syntax for unary expressions. In truth, I'd rather be consistent across the entire codebase, but at a minimum, we should really be consistent across usage within a file:

```suggestion
    if (!m_uwbSessionConnector) {
```

corbin-phipps (changes requested):

windows/devices/uwb/UwbSession.cxx
```diff
@@ -178,69 +178,40 @@ UwbStatus
...
+        PLOG_WARNING << "No associated connector";
+        return UwbStatusGeneric::Failed;
+    }
+
+    auto params = GetApplicationConfigurationParameters({ UwbApplicationConfigurationParameterType::DestinationMacAddresses });
+    if (std::size(params) != 1) {
+        PLOG_ERROR << "did not return the correct number of parameters";
+        return UwbStatusGeneric::Failed;
```
abeltrano, line 189: This is a programmer error and we should flag it immediately. Such code should never get into the wild; ideally, we'll develop a test to fuzz the API and ensure this never happens, but if it does, this exception will flag it so we can fix it:

```suggestion
                throw std::runtime_error("GetApplicationConfigurationParameters() for 1 parameter returned > 1 results. This is a bug!");
```

windows/devices/uwb/UwbSession.cxx
```diff
@@ -178,69 +178,40 @@ UwbStatus
...
+    }
+    auto macAddresses = std::get<std::unordered_set<::uwb::UwbMacAddress>>(params.front().Value);
+    auto [_, inserted] = macAddresses.insert(controleeMacAddress);
+    if (not inserted) {
+        PLOG_INFO << "controleeMacAddress already added, skipping";
+        return UwbStatusGeneric::Ok;
+    }
+    if (macAddresses.size() > MaximumNumberOfControleesInMulticastSession) {
```
abeltrano, line 197: nit: consistency (same for all instances in this file and across this PR):

```suggestion
    if (std::size(macAddresses) > MaximumNumberOfControleesInMulticastSession) {
```

windows/devices/uwb/include/windows/devices/uwb/UwbSession.hxx
```diff
@@ -78,12 +78,21 @@ private:
...
-    GetApplicationConfigurationParametersImpl() override;
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) override;
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
+     */
+    virtual void
+    SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> params) override;
```
abeltrano, line 95: Like other instance, please follow consistent variable naming pattern:

```suggestion
    SetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> uwbApplicationConfigurationParameters) override;
```

abeltrano (changes requested):

lib/uwb/include/uwb/protocols/fira/FiraDevice.hxx
abeltrano: @corbin-phipps When correcting typos, try to point out where exactly the typo is. I usually **bold** or _emphasize_ the part that's incorrect, eg.
Min**u**mumMandatoryMaximumNumberOfControleesInMulticastSession -> Min**i**mumMandatoryMaximumNumberOfControleesInMulticastSession 

But even better, you can use the suggestion feature in GitHub (Ctrl+g) or:
image

```suggestion
constexpr uint8_t MinimumMandatoryMaximumNumberOfControleesInMulticastSession = 8;
```

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -136,12 +136,21 @@ public:
...
      * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
      */
     std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
-    GetApplicationConfigurationParameters();
+    GetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes);
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
```
corbin-phipps, line 150: This comment needs to be updated

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -232,12 +241,21 @@ private:
...
      * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
      */
     virtual std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
-    GetApplicationConfigurationParametersImpl() = 0;
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) = 0;
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
```
corbin-phipps, line 255: Same with this one

windows/devices/uwb/include/windows/devices/uwb/UwbSession.hxx
```diff
@@ -78,12 +78,21 @@ private:
...
      * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
      */
     virtual std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
-    GetApplicationConfigurationParametersImpl() override;
+    GetApplicationConfigurationParametersImpl(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes) override;
+
+    /**
+     * @brief Set the Application Configuration Parameters object
+     *
+     * @param params
```
corbin-phipps, line 92: This one too

corbin-phipps (changes requested):

lib/uwb/include/uwb/UwbSession.hxx
forwardpointer: ok, I think I _finally_ got all of them now 😭, let me know if you spot any that are missing!

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -136,12 +136,21 @@ public:
     SetSessionStatus(const uwb::protocol::fira::UwbSessionStatus& status);
 
     /**
-     * @brief Get the application configuration parameters for this session.
+     * @brief Get the Application Configuration Parameters object
```
abeltrano, line 139: I think the old text should be restored here. This new text looks like the canned doxygen text for a new block.

lib/uwb/include/uwb/UwbSession.hxx
```diff
@@ -136,12 +136,21 @@ public:
...
      *
+     * @param requestedTypes leave this as an empty vector to request all parameters
      * @return std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
      */
     std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
-    GetApplicationConfigurationParameters();
+    GetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes);
+
+    /**
+     * @brief Set the Application Configuration Parameters object
```
abeltrano, line 148: Let's update this text to match the text for the `Get...*` call above (after restoring its old text and mentioned in separate comment).

windows/devices/uwb/UwbSession.cxx
```diff
@@ -178,69 +178,39 @@ UwbStatus
...
+    }
+    auto macAddresses = std::get<std::unordered_set<::uwb::UwbMacAddress>>(params.front().Value);
+    auto [_, inserted] = macAddresses.insert(controleeMacAddress);
+    if (!inserted) {
+        PLOG_INFO << "controleeMacAddress already added, skipping";
+        return UwbStatusGeneric::Ok;
+    }
+    if (std::size(macAddresses) > MaximumNumberOfControleesInMulticastSession) {
+        PLOG_WARNING << "exceeded max number of controlees with numberOfControlees=" << std::size(macAddresses);
+        return UwbStatusGeneric::Rejected;
```
abeltrano, line 198: This seems a bit too generic. Can you consult the spec and see what the expected status should be when # of controlees exceeds the max? I'm thinking either `UwbStatusSession::MulticastListFull`, `UwbStatusGeneric::InvalidRange`, or `UwbStatusGeneric::InvalidParameter`.

corbin-phipps (dismissed):

windows/devices/uwb/UwbSession.cxx
corbin-phipps: From my understanding of the UCI spec, it should be `STATUS_ERROR_MULTICAST_LIST_FULL`: "Indicates when multicast list is full during one to many ranging"

forwardpointer: I chose InvalidRange because NumberOfControlees would be out of the range

forwardpointer: > From my understanding of the UCI spec, it should be `STATUS_ERROR_MULTICAST_LIST_FULL`: "Indicates when multicast list is full during one to many ranging"
> 
> InvalidParameter/Range is used for configuration parameters

I think you're right, while we are setting configuration parameters, STATUS_ERROR_MULTICAST_LIST_FULL fits better here. I had thought that it was specific to SESSION_UPDATE_CONTROLLER_MULTICAST_LIST_NTF, but it looks like that notification has its own table of statuses.

forwardpointer: On the other hand, you might even say that we should let the UWBS return the status, and the role of this function is just to forward that status to the caller. However, the way I implemented it right now is a bit more efficient (but it does guess what the UWBS would return).

abeltrano: I'd fall back to doing what the spec says. If we ever have to get FIRA certified, then we have to do what the spec says. So, I think our options are to either return the spec value here, or, as you suggested @forwardpointer, just pass this through to the driver and let the device provide the status. I'm thinking of preferring the latter since the less the service is involved, the better.

forwardpointer: Thinking about this more, I realize that STATUS_ERROR_ADDRESS_ALREADY_PRESENT doesn't make sense for the UWBS to ever send back if the proper way to update the list of controlees is, as we do here, to send the full list of controlees every time. This means one of two things:
1. STATUS_ERROR_ADDRESS_ALREADY_PRESENT is never sent back as a status when updating controlee list.
2. We misunderstood the proper way to update the list of controlees, and need to redesign this function

abeltrano: Those are good observations. Re-reading UCI spec now, I think it's clear what the expectation is:

* To update controlee list outside of an active session, use `SESSION_SET_APP_CONFIG_CMD`. It sets the complete list every time. Since the session isn't active, this makes sense. This can only return 1 of 3 values: `STATUS_INVALID_PARAM`, `STATUS_INVALID_RANGE`, `STATUS_ERROR_SESSION_ACTIVE` (per spec explicitly listing these 3 only as possible `SESSION_SET_APP_CONFIG_RSP::[AppCfgParam[i]]::Status` field values). 

* To update controlee list during an active session, use `SESSION_UPDATE_CONTROLLER_MULTICAST_LIST_CMD`. This has an action argument that says whether the address specfied should be added or deleted. The response to this command can return a bunch of different values, but I believe if the request made it all the way to the device and basic sanity checks passed, it will return `STATUS_OK`. The final results are provided in a `SESSION_UPDATE_CONTROLLER_MULTICAST_LIST_NTF` notification, where the true status of the update is provided, which includes `STATUS_OK_MULTICAST_LIST_UPDATE`, `STATUS_ERROR_MULTICAST_LIST_FULL`, `STATUS_ERROR_KEY_FETCH_FAIL` or `STATUS_ERROR_SUB_SESSION_ID_NOT_FOUND`. This description in the spec if very explicit as it references the table the values come from (Table 27, which are "Multicast update status codes").

Let's keep it simple for now and remove checks on the max number of controlees from this layer, and just return whatever the connector provides as the result. Motivation is that if we make an assumption, let's localize it to one place. Closer to the driver is better, since information from that layer is more accurate.

forwardpointer: got it. I removed the check on the max number of controlees, but kept the one about the address already present (since it doesn't make sense to send the SESSION_SET_APP_CONFIG_CMD when there is no update).

lib/uwb/UwbSession.cxx
```diff
@@ -131,10 +131,17 @@ UwbSession::InsertPeerImpl(const uwb::UwbMacAddress& peerAddress)
...
     PLOG_VERBOSE << "get application configuration parameters";
-    return GetApplicationConfigurationParametersImpl();
+    return GetApplicationConfigurationParametersImpl(requestedTypes);
+}
+
+void
+UwbSession::SetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter> uwbApplicationConfigurationParameters)
+{
+    PLOG_VERBOSE << "set application configuration parameters";
+    return SetApplicationConfigurationParametersImpl(uwbApplicationConfigurationParameters);
```
abeltrano, line 144: Since a copy is passed to this function, this should really be moved, otherwise a copy will be made passing it to the impl function:

```suggestion
    return SetApplicationConfigurationParametersImpl(std::move(uwbApplicationConfigurationParameters));
```

lib/uwb/UwbSession.cxx
```diff
@@ -131,10 +131,17 @@ UwbSession::InsertPeerImpl(const uwb::UwbMacAddress& peerAddress)
 }
 
 std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameter>
-UwbSession::GetApplicationConfigurationParameters()
+UwbSession::GetApplicationConfigurationParameters(std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> requestedTypes)
 {
     PLOG_VERBOSE << "get application configuration parameters";
-    return GetApplicationConfigurationParametersImpl();
+    return GetApplicationConfigurationParametersImpl(requestedTypes);
```
abeltrano, line 137:
```suggestion
    return GetApplicationConfigurationParametersImpl(std::move(requestedTypes));
```

corbin-phipps (approved):

PR #258, opened 2023-04-27
Add OOB Simulator source to nearobject-framework windows tools

corbin-phipps:
### Type

- [ ] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR adds the Out-of-Band (OOB) Simulator application into the repository. This application roughly simulates an OOB connection, allowing the user to set UWB Session Data options before starting a NearObject Session for UWB ranging with the nocli tool.

This PR is not intended to contain the complete code, just enough to perform the basic functionality. Future PRs will finish and cleanup the application code.

### Technical Details

- Added WinUI 3 application

### Test Results

Ran the application from the Visual Studio solution. Verified that a user can set the required UWB Session Data parameters and the serialized data is written to a file.

Need to verify that the application will run just using the files pushed here to the Git repository.

### Reviewer Focus

- There are a lot of files, but the main code that I touched is in MainWindow.xaml, MainWindow.xaml.cpp/h, Utils.cxx/hxx.
- I don't think we need the Assets folder, but the code won't build without it. Maybe other changes are needed elsewhere.

### Future Work

- The app can supposedly be packaged with MSIX, but I have not tried it. I have only tried running the application from the Visual Studio solution.
- Much work is still needed for things like input validation, error handling, and adding the remaining session data and uwb configuration parameters.
- Need to verify that the serialization into MessagePack is valid and can be reversed on the nocli side.

### Checklist


windows/tools/OobSimulator/App.xaml.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
```
forwardpointer, line 2: Could you run the auto-formatter on this and other files?

forwardpointer (changes requested):

windows/tools/OobSimulator/App.xaml.cpp
corbin-phipps: I don't have the clang-tidy nor clang-format files in this project, but I can add them

abeltrano:
> * I don't think we need the Assets folder, but the code won't build without it. Maybe other changes are needed elsewhere.
The app platform requires this, so we'll need to keep it.

> ### Future Work
> * The app can supposedly be packaged with MSIX, but I have not tried it. I have only tried running the application from the Visual Studio solution.
For packaging, this is much easier to do earlier on in the application lifetime. So, we should not put this off for long. If this will only ever consume existing code and WinRT components, then we might not need packaging at all. However, that will limit what we can do with the app, so I'm partial to ensuring it's packaged correctly.

> * Much work is still needed for things like input validation, error handling, and adding the remaining session data and uwb configuration parameters.
I wouldn't spend too much time on validation right now. We can create a single tracking issue for stuff like this, then just update it with fit-and-finish type items.

windows/tools/OobSimulator/MainWindow.xaml.cpp
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    void MainWindow::SetUwbSessionData(UwbSessionData const& uwbSessionData, std::string const& outputDirectory)
+    {
+        // Convert the input data into JSON
+        nlohmann::json json{};
+        json["sessionId"] = uwbSessionData.sessionId;
+        json["deviceRole"] = static_cast<int>(uwbSessionData.uwbConfiguration.GetDeviceRole().value());
+        json["multiNodeMode"] = static_cast<int>(uwbSessionData.uwbConfiguration.GetMultiNodeMode().value());
+        json["controllerMacAddress"] = uwbSessionData.uwbConfiguration.GetControllerMacAddress().value().ToString();         // Is ToString() the best way to store this?
```
abeltrano, line 35: Great callout, this is important.

Using `ToString()` has some risk because it's not assumed to be stable over time. The library that exposes `UwbSessionData` should expose either serialize/deserialize functions (eg. , say `ToJson()`, `FromJson()`) or some kind of non-member helper code which serializes/deserializes the type. We've used the latter already with `nlohmann` which supports defining `to_json` and `from_json` functions in the hosting namespace to define serialization/deserialization code. See [`UwbMacAddressJsonSerializer.cxx`](https://github.com/microsoft/nearobject-framework/blob/develop/lib/nearobject/serialization/UwbMacAddressJsonSerializer.cxx) for an example. This way the library itself provides a strong contract that guarantees stability over time which consumers can use and rely on. The reason we preferred the non-member approach is that it separates the concerns of the functionality of the class and the serialization of the class. There's really no reason the class functionality need be concerned about the serialization requirements outside of it.

Coincidentally, most of the types you need here are either primitive or already have `nlohmann` serializers defined. So, I suggest adding an `nlohmann` json serializer for `UwbSessionData`.

windows/tools/OobSimulator/MainWindow.xaml
```diff
@@ -0,0 +1,66 @@
...
+<Window
+    x:Class="OobSimulator.MainWindow"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:OobSimulator"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <Grid>
```
abeltrano, line 13: Have you considered some kind of top-level navigational element, like a [NavigationView](https://learn.microsoft.com/en-us/windows/apps/design/controls/navigationview)? If this app is ever going to have more than 1 feature (I'd guess yes), then it's important to plan for the overall navigation structure now. I'd lean towards adding one here than not.

abeltrano (changes requested):
This is really cool. It seems like it will do the trick as it is, but there are some important foundational bits I'd like to establish now/early before we start building too much on top of it.

windows/tools/OobSimulator/MainWindow.xaml
corbin-phipps: That's a good idea, I can add one

windows/tools/OobSimulator/App.xaml.cpp
forwardpointer: .clang-tidy and .clang-format are both present in the top-level of nearobject-framework, so I believe it is possible to right-click this file in VSCode and run the auto-formatter.

corbin-phipps: Oh I see. I ended up adding them in and running the formatter, but I had to add "clang-format off"/"clang-format on" in a couple of places due to the precompiled headers that are automatically part of this project

corbin-phipps: But it looks like it didn't catch everything for some reason, so I need to go back and fix some things

forwardpointer (approved):

windows/tools/OobSimulator/MainWindow.xaml
```diff
@@ -0,0 +1,66 @@
...
+        </StackPanel>
+
+        <StackPanel Grid.Row="5" Orientation="Horizontal" Margin="10">
+            <TextBlock Text="Controlee Short MAC Address: " VerticalAlignment="Center" />
+            <TextBox x:Name="ControleeShortMacAddressTextBox" Margin="10" VerticalAlignment="Center" Text="67:89" />
+        </StackPanel>
+
+        <StackPanel Grid.Row="6" Orientation="Horizontal" Margin="10">
+            <TextBlock Text="Output Directory: " VerticalAlignment="Center" />
+            <TextBox x:Name="OutputDirectoryTextBox" Margin="10" VerticalAlignment="Center" Text="C:/..." />
```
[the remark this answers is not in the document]
abeltrano: There's so many controls, hard to know they even exist sometimes :)

windows/tools/OobSimulator/MainWindow.xaml.cpp
abeltrano: We did a lot of the service work during the original design. It was mostly just quick and dirty stuff for prototyping and shaping, but a lot of it turned out to be useful as it was.

windows/tools/OobSimulator/MainWindow.xaml
abeltrano: As discussed offline, let's leave to separate PR.

windows/tools/OobSimulator/MainWindow.xaml
```diff
@@ -0,0 +1,66 @@
...
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+
+        <TextBlock Grid.Row="0" Text="Set Session Data" FontSize="24" FontWeight="Bold" Margin="10" />
+
+        <StackPanel Grid.Row="1" Orientation="Horizontal" Margin="10">
+            <TextBlock Text="UWB Session ID: " VerticalAlignment="Center" />
+            <TextBox x:Name="UwbSessionIdTextBox" Margin="10" VerticalAlignment="Center" Text="1234" />
```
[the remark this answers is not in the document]
abeltrano: As discussed offline, let's leave to separate PR.

abeltrano (approved):

abeltrano (approved):

PR #263, opened 2023-05-04
Decouple logic from UI in OobSimulator app

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR moves away from a tightly coupled UI and application logic to a decoupled approach using MVVM (Model-View-ViewModel).

### Technical Details

- Added UwbSessionData runtime class as the Model.
- Added OobSimulatorViewModel runtime class as the ViewModel.
- Updated MainWindow (View) to bind elements to properties in the ViewModel.
- Added BindableBase and DelegateCommand runtime classes, which are needed for binding the button to a function that runs in the ViewModel.
- Updated UwbConfigurationJsonSerializer to check if std::optional values actually have values in order to prevent exceptions.

### Test Results

Verified that data from the text boxes and radio buttons are serialized and written to a file, just like before the design change.

### Reviewer Focus

There is a lot here, but following MVVM, the important UwbSessionData is stored in the Model (UwbSessionData.idl); the UI is stored in the View (MainWindow.xaml), and the application logic that bridges the gaps is stored in the ViewModel (OobSimulatorViewModel.cpp).

For reference, this page was very helpful in my development: https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/binding-property as well as this: https://github.com/johanlindfors/MvvmCppWinRT/blob/master/App.xaml

### Future Work

- There is currently no binding for the `RadioButtons`, so only the default values can be set for `DeviceRole` and `MultiNodeMode`. The reason for the missing binding is because an implementation of `IValueConverter` is needed to bind the `IsChecked` property of a `RadioButton`, but there is seemingly no way to add a converter to a top-level `Window` UI element, which is what the Blank WinUI 3 app created instead of a `Page`, which is much more common and used in UWP apps. I have the code for one of the converters locally, but I could not figure out how to use it!
- Related to the previous item, we should change `Window` to `Page` and allow for multiple `Page`s in the UI, with some sort of navigation. I did not get to this in this PR because just the MVVM changes took A TON of time due to various WinRT issues...
- Also related to previous items, once `Window` is changed to `Page`, we can remove `Globals.h` and access the current `Window` in a different way instead of storing the handle globally from `MainWindow.xaml.cpp`.

### Checklist


windows/tools/OobSimulator/BindableBase.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
forwardpointer: nit: Here, and in other files, I think it would be nice to use \#ifdef guards instead of pragma, just like the reset of the nearobject-framework

windows/tools/OobSimulator/DelegateCommand.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    m_execute = execute;
+    m_canExecute = canExecute;
+}
+
+void
+DelegateCommand::Execute(IInspectable const& parameter)
+{
+    if (m_execute != nullptr) {
```
forwardpointer: nit: Since std::function can only be checked for equality with nullptr (since it's not a pointer type), I think that it would be better to use the std::function::operator bool() check here

```suggestion
    if (m_execute) {

File: windows/tools/OobSimulator/DelegateCommand.h [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+    void
+    CanExecuteChanged(winrt::event_token const& token);
+
+    void
+    RaiseCanExecuteChanged();
+
+private:
+    std::function<void(winrt::Windows::Foundation::IInspectable)> m_execute;
+    std::function<bool(winrt::Windows::Foundation::IInspectable)> m_canExecute;
+    bool m_lastCanExecute = true;
```
forwardpointer, line 31: nit: I think it would make more sense to rename m_lastCanExecute and m_canExecute to more clearly indicate that one is a bool and the other is a function, otherwise they seem like they are the same type at first glance.

windows/tools/OobSimulator/OobSimulatorViewModel.cpp
```diff
@@ -0,0 +1,107 @@
...
+OobSimulatorViewModel::SetUwbSessionDataCommand()
+{
+    return m_setUwbSessionData;
+}
+
+winrt::Windows::Foundation::IAsyncAction
+OobSimulatorViewModel::SetUwbSessionData()
+{
+    uwb::protocol::fira::UwbSessionData uwbSessionData;
+    uwb::protocol::fira::UwbConfiguration::Builder builder{};
```
forwardpointer, line 45: nit: Try to keep the way you call the default constructor consistent (either use {} or not)

windows/tools/OobSimulator/OobSimulatorViewModel.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    // UWB Session ID
+    uwbSessionData.sessionId = HstringToUint32(m_uwbSessionData.SessionId());
+
+    // Convert OobSimulator::DeviceRole to uwb::protocol::fira::DeviceRole
+    auto oobSimulatorDeviceRole = m_uwbSessionData.DeviceRole();
```
forwardpointer: It might make sense to put the conversion routine into its own function

forwardpointer (dismissed):
I'm not very familiar with UI design in cpp but what you have so far looks good. I just left comments on some nits I had

windows/tools/OobSimulator/DelegateCommand.cpp
corbin-phipps: This is a great suggestion, thanks

windows/tools/OobSimulator/DelegateCommand.h
corbin-phipps: I changed m_execute and m_canExecute (the functions) to m_executeDelegate and m_canExecuteDelegate to differentiate between the functions and the bool field (m_lastCanExecute)

forwardpointer (approved):
This is good. One last nit I have is that I would like to see the h and cpp files renamed to hxx and cxx files to maintain consistency with the rest of the repo.

corbin-phipps:
> This is good. One last nit I have is that I would like to see the h and cpp files renamed to hxx and cxx files to maintain consistency with the rest of the repo.

Sure, I can change them. The .h and .cpp extensions are just the defaults when adding a new WinRT runtime class implementation (I would add a .idl file, build that to generate stubs, which are .h and .cpp)

EDIT: I'm actually having some problems building with the changed file names, so I will leave this to another PR

PR #269, opened 2023-05-12
Implement remaining OOB serialization functions

abeltrano:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* Allow complete serialization of `UwbSessionData` for use in OOB.

### Technical Details

* Implement `[To|From]DataObject` functions for all objects starting at the root of the `UwbSessionData` tree.
* Update `TlvSet*` functions to take values instead of refs since copies were being made anyway, so the interface taking value is more clear about the behavior.
* Remove non-existent channels from `UwbRegulatoryInformation` and fix tag values.
* Add missing `MaximumTransmissionPower` property to `UwbRegulatoryInformation`.
* Add missing `UwbConfigurationAvailable` property to `UwbSessionData`.

### Test Results

* All unit tests pass on both Linux and Windows.

### Reviewer Focus

Nothing specific.

### Future Work

None

### Checklist


lib/uwb/protocols/fira/UwbRegulatoryInformation.cxx
```diff
@@ -0,0 +1,181 @@
...
+                .SetValue(MaximumTransmissionPower.at(Channel::C14))
+                .Build())
+        .Build();
+
+    return tlvBerResult;
+}
+
+/* static */
+UwbRegulatoryInformation
+UwbRegulatoryInformation::FromDataObject(const encoding::TlvBer& tlvBer)
```
forwardpointer, line 99: Should all of these FromDataObject functions signify to the caller if it couldn't decode the tlv? Right now it looks like it returns a default-constructed object if it couldn't decode.

abeltrano: That's a good point. The cleanest is probably to `throw` if decoding failed. We'll have to think of the best way to determine when failure actually occurs. These functions seem to need more logic at the end to determine if all required tags were present and decoded properly and if not, then indicate the failure to the caller in some way.

abeltrano: I added a basic validation check and if not met, throw `UwbException(UwbStatusGeneric::SyntaxError)`. I think this will be good enough for now.

PR #271, opened 2023-05-15
Enable TLV-BER in OobSimulator

corbin-phipps:
### Type

- [ ] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

This PR enables the TLV-BER encoding of UwbSessionData in the OobSimulator app. Users of this app can choose to save the encoding as TLV-BER or as MessagePack.

### Technical Details

- Added SessionDataEncoding enum and property to View (UwbSessionData).
- Added radio button options for the session data encoding.
- Added code to save the TLV-BER encoded session data as a ".tlvber" file.

### Test Results

Verified the ability to choose between TLV-BER and MessagePack options, as well as the ability to save the encodings as their respective file types.

### Reviewer Focus

None.

### Future Work

I used a simple event handler in the XAML code-behind file to handle the radio button selection, but this should eventually be replaced to use IValueConverter (along with the other radio buttons). I was previously unsuccessful with using the converters, so this new feature takes the simpler, non-MVVM approach.

### Checklist

### DIFF
--- a/windows/tools/OobSimulator/UwbSessionData.cpp
+++ b/windows/tools/OobSimulator/UwbSessionData.cpp
@@ -71,6 +71,19 @@ UwbSessionData::ControleeShortMacAddress(hstring const& value)
         m_propertyChanged(*this, winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventArgs{ L"Controlee Short MAC Address" });
     }
 }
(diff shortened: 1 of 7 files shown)


abeltrano (approved):
Nice!

PR #280, opened 2023-05-23
Add DeviceType parameter to CreateSession

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [x] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

The `DeviceType` parameter is determined by the application utilizing UWB (e.g. `HandleStartRanging()` in nocli takes a `DeviceType` as an argument) and is currently never set in `UwbSession`. Because it is accessed by the WinRT service APIs through `GetDeviceType()`, this parameter must be set, otherwise `GetDeviceType()` will always return the default value. It makes most sense to me to set the `DeviceType` when constructing a session rather than setting it after a session is already constructed.

### Technical Details

- Added `DeviceType` as a parameter to `CreateSession` and `CreateSessionImpl`.

### Test Results

Compile-tested only.

### Reviewer Focus

None.

### Future Work

The WinRT API code that calls `CreateSession` will need to be updated to pass in the `DeviceType`, which comes from the application.

### Checklist

### DIFF
--- a/tools/cli/NearObjectCliHandler.cxx
+++ b/tools/cli/NearObjectCliHandler.cxx
@@ -37,7 +37,17 @@ void
 NearObjectCliHandler::HandleDriverStartRanging(std::shared_ptr<uwb::UwbDevice> uwbDevice, const UwbRangingParameters& rangingParameters) noexcept
 try {
     auto controlFlowContext = (m_parent != nullptr) ? m_parent->GetControlFlowContext() : nullptr;
-    auto session = uwbDevice->CreateSession(rangingParameters.SessionId, m_sessionEventCallbacks);
+
+    auto deviceType = uwb::protocol::fira::DeviceType::Controller;
+    auto parameters = rangingParameters.ApplicationConfigurationParameters;
+    auto it = std::find_if(std::cbegin(parameters), std::cend(parameters), [](const UwbApplicationConfigurationParameter& parameter) {
+        return std::get_if<uwb::protocol::fira::DeviceType>(&parameter.Value) != nullptr;
+    });
+    if (it != std::cend(parameters)) {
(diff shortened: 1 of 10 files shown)


forwardpointer (dismissed):

tools/cli/NearObjectCliHandler.cxx
```diff
@@ -37,7 +37,17 @@ void
...
     auto controlFlowContext = (m_parent != nullptr) ? m_parent->GetControlFlowContext() : nullptr;
-    auto session = uwbDevice->CreateSession(rangingParameters.SessionId, m_sessionEventCallbacks);
+
+    auto deviceType = uwb::protocol::fira::DeviceType::Controller;
+    auto parameters = rangingParameters.ApplicationConfigurationParameters;
+    auto it = std::find_if(std::cbegin(parameters), std::cend(parameters), [](const UwbApplicationConfigurationParameter& parameter) {
+        return std::get_if<uwb::protocol::fira::DeviceType>(&parameter.Value) != nullptr;
+    });
+    if (it != std::cend(parameters)) {
+        deviceType = *(std::get_if<uwb::protocol::fira::DeviceType>(&it->Value));
```
abeltrano: nit: the concrete type is known, so `std::get` should be used instead:

```suggestion
        deviceType = std::get<uwb::protocol::fira::DeviceType>(&it->Value);

```

abeltrano (dismissed):

tools/cli/NearObjectCliHandler.cxx
corbin-phipps: Good point, I'll make the change

abeltrano (approved):

PR #281, opened 2023-05-30
Always send at least sizeof(UWB_NOTIFICATION_DATA) bytes for notifications

corbin-phipps:
### Type

- [x] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Currently, the handling of notifications is failing the UWB CX validation test. This happens because the CX requires a buffer of at least `sizeof(UWB_NOTIFICATION_DATA)` (0x50) bytes, but `lpBytesReturned` of `DeviceIoControl` returns fewer bytes. The reason for this is because `UWB_NOTIFICATION_DATA` has a union member, which causes `sizeof` to account for the _largest_ member in the union. A specific notification of the non-largest type may actually use fewer bytes than what `sizeof` calculates for the union.

Therefore, the notification handling should always send at least `sizeof(UWB_NOTIFICATION_DATA)` bytes, not just the number of bytes actually used for a specific notification.

### Technical Details

- Changed the notification handling to resize the buffer to the maximum of `bytesRequired` and `sizeof(UWB_NOTIFICATION_DATA)`.

### Test Results

nocli output displays expected values, eliminating the unexpected "insufficient buffer" messages from before.

### Reviewer Focus

None.

### Future Work

None.

### Checklist

### DIFF
--- a/windows/devices/uwb/UwbConnector.cxx
+++ b/windows/devices/uwb/UwbConnector.cxx
@@ -685,11 +685,12 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
     auto handleDriver = m_notificationHandleDriver;
 
     while (!stopToken.stop_requested()) {
-        DWORD bytesRequired = sizeof(UWB_NOTIFICATION_DATA);
+        DWORD minimumNotificationSize = sizeof(UWB_NOTIFICATION_DATA);
+        DWORD bytesRequired = minimumNotificationSize;
         std::vector<uint8_t> uwbNotificationDataBuffer{};
         m_notificationOverlapped = {};
         for (const auto i : std::ranges::iota_view{ 0, 2 }) {
-            uwbNotificationDataBuffer.resize(bytesRequired);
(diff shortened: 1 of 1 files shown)


windows/devices/uwb/UwbConnector.cxx
```diff
@@ -685,11 +685,12 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
     auto handleDriver = m_notificationHandleDriver;
 
     while (!stopToken.stop_requested()) {
-        DWORD bytesRequired = sizeof(UWB_NOTIFICATION_DATA);
+        DWORD minimumNotificationSize = sizeof(UWB_NOTIFICATION_DATA);
```
abeltrano:
```suggestion
        static constexpr DWORD minimumNotificationSize = sizeof(UWB_NOTIFICATION_DATA);
```

abeltrano (dismissed):

abeltrano (approved):

forwardpointer (approved):

PR #290, opened 2023-06-07
add explicit set tag functionfor uint16_t and size_t to handle endianness

forwardpointer:
### Type

- [x] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

* add a function to set the tag for uint16_t and size_t to handle endianness, since the TlvBer needs a big endian tag
* add a high level round trip test for tlvber that uses the uwbsessiondata

### Test Results

tests pass locally 

### Reviewer Focus

This requires the user to explicitly tell the compiler which SetTag function to use.

### Future Work

Maybe put the test somewhere else since it uses components outside of TlvBer

### Checklist

### DIFF
--- a/lib/shared/tlv/TlvBer.cxx
+++ b/lib/shared/tlv/TlvBer.cxx
@@ -1,6 +1,7 @@
 
 #include <algorithm>
 #include <bit>
+#include <cstring>
 #include <iterator>
 #include <optional>
 #include <utility>
--- a/lib/shared/tlv/TlvBer.cxx
+++ b/lib/shared/tlv/TlvBer.cxx
@@ -232,6 +232,44 @@ TlvBer::Builder::SetTag(uint8_t tag)
     return *this;
 }
 
+TlvBer::Builder&
+TlvBer::Builder::SetTag(uint16_t tag)
+{
+    std::array<uint8_t, sizeof tag> tagData;
+    std::memcpy(&tagData, &tag, sizeof tag);
+
+    if (std::endian::native == std::endian::big) {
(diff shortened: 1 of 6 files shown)


lib/shared/tlv/TlvBer.cxx
```diff
@@ -232,6 +232,44 @@ TlvBer::Builder::SetTag(uint8_t tag)
...
+        std::transform(std::rbegin(tagData), std::rend(tagData), std::begin(tagDataRev), [](auto&& b) {
+            return b;
+        });
+        TlvBer::Builder::SetTag(tagDataRev);
+        return *this;
+    }
+}
+
+TlvBer::Builder&
+TlvBer::Builder::SetTag(size_t tag)
```
abeltrano, line 256: Is this version actually needed anywhere? Alternatively, since unsigned types are implicitly convertible to `std::size_t`, could we only use this version and remove the `uint16_t` version?

abeltrano (dismissed):

lib/shared/tlv/TlvBer.cxx
forwardpointer: I don't think so. 
When I just had the uint16_t version the compiler complained about some file needing the size_t version. 
Regarding the other direction, size_t is always larger than uint16_t so if we enter a uint16_t as use it as a size_t, the resultant tag would have leading zero bytes, which is undesirable.

PR #306, opened 2023-06-14
Add check for handle validity in HandleNotifications

corbin-phipps:
### Type

- [x] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

In a rare situation while running a UWB ranging session with nocli, the underlying driver will crash (ERROR_DRIVER_PROCESS_TERMINATED) but the notification handle from UwbConnector is still open, so nocli will continuously print errors and will not exit on Ctrl+C. This happens because the notification loop runs until a stop is requested (`stopToken.stop_requested()`), but in this case, no stop has been requested.

### Technical Details

- In `HandleNotifications()`, added a check for the validity of the driver handle in addition to the requested stop of the `std::stop_source`.

### Test Results

Compile-tested. No regression in ranging scenario with CX + TestClient driver.

This issue is almost impossible to repro, but it was observed once in real hardware integration testing.

### Reviewer Focus

Does this change correctly fix the problem? Another option is `nullptr != handleDriver.get()`. I had trouble finding any information on the `wil::shared_hfile` object, so I'm assuming that the `is_valid()` function accomplishes the same thing.

### Future Work

None.

### Checklist


windows/devices/uwb/UwbConnector.cxx
```diff
@@ -690,7 +690,7 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
 
     auto handleDriver = m_notificationHandleDriver;
 
-    while (!stopToken.stop_requested()) {
+    while (!stopToken.stop_requested() && handleDriver.is_valid()) {
```
abeltrano, line 693: `is_valid()` on a handle will tell you that the handle value is `!= INVALID_HANDLE_VALUE`. So, it's a check that's only valuable after the handle is opened (and this is already done for this handle in [`OpenDriverHandle`](https://github.com/microsoft/nearobject-framework/blob/c2b7c367acbb12749336e2f5de865dc7eb45ed75/windows/devices/util/include/windows/devices/DeviceHandle.hxx#L60), but with an explicit check instead of using `is_valid()`).

I believe the only way to detect the condition mentioned in the description is to handle `ERROR_DRIVER_PROCESS_TERMINATED` directly from the IOCTL, and in that case, invalidate the handle by calling `.reset()` on it (which will cause `is_valid()` to return false) and keep the newly added `is_valid()` check here.

abeltrano (changes requested):

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: Ah I see, I wasn't sure if the driver crash would cause the handle to automatically become invalid or not. I can handle the `ERROR_DRIVER_PROCESS_TERMINATED` error like you said

windows/devices/uwb/UwbConnector.cxx
```diff
@@ -736,11 +740,15 @@ UwbConnector::HandleNotifications(std::stop_token stopToken)
...
-                    LOG_VERBOSE << logPrefix << "insufficient buffer, " << bytesRequired << " bytes required, current size " << std::size(uwbNotificationDataBuffer) << " bytes";
+                    PLOG_VERBOSE << logPrefix << "insufficient buffer, " << bytesRequired << " bytes required, current size " << std::size(uwbNotificationDataBuffer) << " bytes";
                     const UWB_NOTIFICATION_DATA& notificationDataPartial = *reinterpret_cast<UWB_NOTIFICATION_DATA*>(std::data(uwbNotificationDataBuffer));
                     bytesRequired = std::max(notificationDataPartial.size, static_cast<uint32_t>(minimumNotificationSize));
                     // Attempt to retry the ioctl with the appropriate buffer size, which is now held in bytesRequired.
                     continue;
+                } else if (lastError == ERROR_DRIVER_PROCESS_TERMINATED) {
+                    PLOG_ERROR << logPrefix << "driver process terminated, hr=" << std::showbase << std::hex << hr;
+                    handleDriver.reset();
+                    break; // for({1,2})
```
abeltrano, line 751: These comments are referring to the old loop exit condition. Since there's only 1 for loop here, it might be fine to just say `// for()`.

```suggestion
                    break; // for()
```

We might just do this in another PR though, so feel free to close this without addressing it, I just wanted to point it out. They're both wrong and inconsistent ({0,1}, {0,2}, I think both done by me), so we should fix it.

abeltrano (approved):

windows/devices/uwb/UwbConnector.cxx
corbin-phipps: Sounds good to leave it for another PR - I just copied those comments without thinking about them and I agree with your suggestion

PR #308, opened 2023-06-15
Expose OOB data bytes through UwbSession

corbin-phipps:
### Type

- [ ] Bug fix
- [x] Feature addition
- [ ] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

`UwbSession` currently does not have a way of exposing `UwbSessionData` from it. This PR exposes that data in its Data Object format as a byte buffer.

### Technical Details

- Added OOB conversion function: `GetUwbSessionData`.
- Added public `GetOobDataObject` function in `UwbSession`.

### Test Results

Compile-tested only.

### Reviewer Focus

None.

### Future Work

The UCI parameters and the OOB parameters do not match up 1:1, so there are quite a few TODOs in the code for these situations.

The classic "controlee short mac address" problem shows up here in the conversion function, so the code currently assumes short mac addresses and only 1 destination mac address.

### Checklist

### DIFF
--- a/lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
@@ -3,7 +3,7 @@
 #define UWB_OOB_CONVERSIONS_HXX
 
 #include <uwb/protocols/fira/FiraDevice.hxx>
-#include <uwb/protocols/fira/UwbConfiguration.hxx>
+#include <uwb/protocols/fira/UwbSessionData.hxx>
 
 namespace uwb::protocol::fira
 {
@@ -14,6 +14,14 @@ namespace uwb::protocol::fira
  */
 std::vector<UwbApplicationConfigurationParameter>
 GetUciConfigParams(const UwbConfiguration& uwbConfiguration, DeviceType deviceType);
+
+/**
+ * @brief Converts the given UCI application configuration parameters to the OOB UwbSessionData equivalent
+ * 
+ * @return UwbSessionData
+*/
+UwbSessionData
+GetUwbSessionData(std::vector<UwbApplicationConfigurationParameter> applicationConfigurationParameters);
 
 } // namespace uwb::protocol::fira
 
--- a/lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
@@ -17,9 +17,9 @@ GetUciConfigParams(const UwbConfiguration& uwbConfiguration, DeviceType deviceTy
 
 /**
  * @brief Converts the given UCI application configuration parameters to the OOB UwbSessionData equivalent
- * 
+ *
  * @return UwbSessionData
-*/
+ */
 UwbSessionData
 GetUwbSessionData(std::vector<UwbApplicationConfigurationParameter> applicationConfigurationParameters);
 
--- a/lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
+++ b/lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
@@ -10,6 +10,8 @@ namespace uwb::protocol::fira
 /**
  * @brief Converts the config params given by OOB to config params that UCI needs
  *
+ * @param uwbConfiguration The UWB configuration data used to generate the UCI configuration parameter list
+ * @param deviceType The type of device (Controller/Controlee)
  * @return std::vector<UwbApplicationConfigurationParameter>
  */
 std::vector<UwbApplicationConfigurationParameter>
@@ -18,6 +20,7 @@ GetUciConfigParams(const UwbConfiguration& uwbConfiguration, DeviceType deviceTy
 /**
  * @brief Converts the given UCI application configuration parameters to the OOB UwbSessionData equivalent
  *
+ * @param applicationConfigurationParameters The UCI application configuration parameters used to construct UwbSessionData
  * @return UwbSessionData
  */
 UwbSessionData
--- a/lib/uwb/protocols/fira/UwbOobConversions.cxx
+++ b/lib/uwb/protocols/fira/UwbOobConversions.cxx
@@ -1,4 +1,6 @@
 
+#include <magic_enum.hpp>
+#include <plog/Log.h>
 #include <uwb/protocols/fira/UwbConfigurationBuilder.hxx>
 #include <uwb/protocols/fira/UwbOobConversions.hxx>
 #include <uwb/protocols/fira/UwbSessionData.hxx>
@@ -164,9 +166,6 @@ uwb::protocol::fira::GetUwbSessionData(std::vector<UwbApplicationConfigurationPa
         case UwbApplicationConfigurationParameterType::DeviceType:
             deviceType = std::get<DeviceType>(parameterValue);
             break;
-        case UwbApplicationConfigurationParameterType::RangingRoundUsage:
-            // TODO
-            break;
         case UwbApplicationConfigurationParameterType::StsConfiguration:
             builder.SetStsConfiguration(std::get<StsConfiguration>(parameterValue));
             break;
@@ -175,9 +174,6 @@ uwb::protocol::fira::GetUwbSessionData(std::vector<UwbApplicationConfigurationPa
             break;
         case UwbApplicationConfigurationParameterType::ChannelNumber:
             builder.SetChannel(std::get<Channel>(parameterValue));
-            break;
-        case UwbApplicationConfigurationParameterType::NumberOfControlees:
-            // TODO
             break;
         case UwbApplicationConfigurationParameterType::DeviceMacAddress:
             // In case this parameter comes before DeviceType, store and set correct parameter later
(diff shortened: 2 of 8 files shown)


lib/uwb/include/uwb/protocols/fira/UwbOobConversions.hxx
```diff
@@ -15,6 +15,14 @@ namespace uwb::protocol::fira
 std::vector<UwbApplicationConfigurationParameter>
 GetUciConfigParams(const UwbConfiguration& uwbConfiguration, DeviceType deviceType);
 
+/**
+ * @brief Converts the given UCI application configuration parameters to the OOB UwbSessionData equivalent
+ *
+ * @return UwbSessionData
+ */
+UwbSessionData
+GetUwbSessionData(std::vector<UwbApplicationConfigurationParameter> applicationConfigurationParameters);
```
abeltrano: nit: function docs missing description of `applicationConfigurationParameters` argument.

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -142,4 +144,192 @@ uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& uwbConfiguration
...
+        case UwbApplicationConfigurationParameterType::DeviceMacAddress:
+            // In case this parameter comes before DeviceType, store and set correct parameter later
+            deviceMacAddress = std::get<UwbMacAddress>(parameterValue);
+            builder.SetMacAddressController(deviceMacAddress);
+            break;
+        case UwbApplicationConfigurationParameterType::DestinationMacAddresses:
+            // In case this parameter comes before DeviceType, store and set correct parameter later
+            destinationMacAddresses = std::get<std::unordered_set<UwbMacAddress>>(parameterValue);
+            // TODO: Do this correctly
+            builder.SetMacAddressControleeShort(*destinationMacAddresses.begin());
```
abeltrano: style consistency:

```suggestion
            builder.SetMacAddressControleeShort(std::begin(*destinationMacAddresses));
```

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -142,4 +144,192 @@ uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& uwbConfiguration
...
+            break;
+        case UwbApplicationConfigurationParameterType::DeviceMacAddress:
+            // In case this parameter comes before DeviceType, store and set correct parameter later
+            deviceMacAddress = std::get<UwbMacAddress>(parameterValue);
+            builder.SetMacAddressController(deviceMacAddress);
+            break;
+        case UwbApplicationConfigurationParameterType::DestinationMacAddresses:
+            // In case this parameter comes before DeviceType, store and set correct parameter later
+            destinationMacAddresses = std::get<std::unordered_set<UwbMacAddress>>(parameterValue);
+            // TODO: Do this correctly
```
abeltrano, line 186: definitely fine for now 👍

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -142,4 +144,192 @@ uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& uwbConfiguration
...
+            break;
+        case UwbApplicationConfigurationParameterType::BprfPhrDataRate:
+            // TODO
+            break;
+        case UwbApplicationConfigurationParameterType::MaxNumberOfMeasurements:
+            // TODO
+            break;
+        case UwbApplicationConfigurationParameterType::StsLength:
+            // TODO
+            break;
```
abeltrano: We should at least log the ones we're not processing so someone can discover that. It might be worth grouping all of the ones with `TODO`s, letting them fall-through to the last unsupported case (make sure to annotate each case with [`[[fallthrough]]`](https://en.cppreference.com/w/cpp/language/attributes/fallthrough)), and then log a warning in that case.
```suggestion
        // move others with TODOs here
        case UwbApplicationConfigurationParameterType::BprfPhrDataRate: [[fallthrough]];
        case UwbApplicationConfigurationParameterType::MaxNumberOfMeasurements: [[fallthrough]];
        case UwbApplicationConfigurationParameterType::StsLength:
            `PLOG_WARNING << "parameter type " << magic_enum::enum_name(parameterType) << " not implemented; ignoring";`
            break;
```

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -142,4 +144,192 @@ uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& uwbConfiguration
...
+            // TODO: Log error
+            break;
+        }
+    }
+
+    // TODO: This assumes short mac addresses and only 1 address in the destination set.
+    //       Need to figure out how to do this correctly
+    if (deviceType == DeviceType::Controller) {
+        builder.SetMacAddressController(deviceMacAddress);
+        builder.SetMacAddressControleeShort(*destinationMacAddresses.begin());
```
abeltrano: style consistency:
```suggestion
        builder.SetMacAddressControleeShort(std::begin(*destinationMacAddresses));
```

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -142,4 +144,192 @@ uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& uwbConfiguration
...
+    } else { // DeviceType::Controlee
+        builder.SetMacAddressController(*destinationMacAddresses.begin());
```
abeltrano: style consistency:
```suggestion
        builder.SetMacAddressController(std::begin(*destinationMacAddresses));
```

lib/uwb/protocols/fira/UwbOobConversions.cxx
```diff
@@ -142,4 +144,192 @@ uwb::protocol::fira::GetUciConfigParams(const UwbConfiguration& uwbConfiguration
...
+            break;
+        case UwbApplicationConfigurationParameterType::ResponderSlotIndex:
+            // TODO
+            break;
+        case UwbApplicationConfigurationParameterType::PrfMode:
+            if (std::get<PrfModeDetailed>(parameterValue) == PrfModeDetailed::Bprf62MHz) {
+                builder.SetPrfMode(PrfMode::Bprf);
+            } else { // PrfModeDetailed::Hprf124MHz || PrfModeDetailed::Hprf249MHz
+                builder.SetPrfMode(PrfMode::Hprf);
+            }
```
abeltrano: Would rather see the explicit check here so future changes (however unlikely) don't cause this to stop working if a newly added enum value is used.
```suggestion
            const auto prfModeValue = std::get<PrfModeDetailed>(parameterValue);
            switch (prfModeValue) {
            case PrfModeDetailed::Bprf62MHz:
                builder.SetPrfMode(PrfMode::Bprf);
                break;
            case PrfModeDetailed::Hprf124MHz: [[ fallthrough ]];
            case PrfModeDetailed::Hprf249MHz:
                builder.SetPrfMode(PrfMode::Hprf);
                break;
            default:
                PLOG_WARNING << "unsupported PRF Mode value " << magic_enum::enum_name(prfModeValue) << " specified; ignoring";
                break;
            }
```

windows/devices/uwb/UwbSession.cxx
```diff
@@ -295,3 +296,15 @@ UwbSession::DestroyImpl()
         throw e;
     }
 }
+
+std::vector<uint8_t>
+UwbSession::GetOobDataObjectImpl()
+{
+    auto applicationConfigurationParameters = GetApplicationConfigurationParameters({});
```
abeltrano: Specifying none (empty vector) has a special meaning to request all parameters, right? If so, we should define a constant for this to make it more clear in the code. Probably makes sense to define this at/close/above the prototype for `UwbSession::GetApplicationConfigurationParameters()` in UwbSession.hxx. Eg. (example defines it here instead for simplicity):

```suggestion
    static constexpr std::vector<::uwb::protocol::fira::UwbApplicationConfigurationParameterType> AllParameters = {};
    auto applicationConfigurationParameters = GetApplicationConfigurationParameters(AllParameters);
```

abeltrano (dismissed):
A few comments here related to style and fit-and-finish, but this looks good overall.

lib/uwb/protocols/fira/UwbOobConversions.cxx
corbin-phipps: Oh that's a good idea, I'll do that

windows/devices/uwb/UwbSession.cxx
corbin-phipps: Yes that's correct. I think the empty vector as a param is used somewhere else in the code (nocli maybe?), so it may make sense to define the constant at a broader level. I'll double check though

windows/devices/uwb/UwbSession.cxx
```diff
@@ -295,3 +296,15 @@ UwbSession::DestroyImpl()
...
     }
 }
+
+std::vector<uint8_t>
+UwbSession::GetOobDataObjectImpl()
+{
+    auto applicationConfigurationParameters = GetApplicationConfigurationParameters(AllParameters);
+    auto uwbSessionData = GetUwbSessionData(applicationConfigurationParameters);
+
+    uwbSessionData.sessionId = GetId();
```
abeltrano, line 306: We should figure out an appropriate default version for `uwbSessionData.sessionDataVersion`. The spec says:
> This two byte field should be interpreted as two unsigned integer numbers representing the major and minor version.

But I don't know where this version comes from. Right now, we default it to 0, which is almost certainly wrong. Their example was `0x0101` for v1.1, so I'm inclined to use that. I'll make a separate PR to do this.

abeltrano (dismissed):

windows/devices/uwb/UwbSession.cxx
abeltrano: I made the change here when I fixed the Linux build.

PR #309, opened 2023-06-19
Move session status changed callback from device to session callbacks

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Breaking change
- [ ] Non-functional change
- [ ] Documentation
- [ ] Infrastructure

### Goals

Simplify registering for session callbacks. The existing logic had the session status changed callback in the device event callbacks. The reasoning was that the `UwbDevice` would probably need to manage sessions even if the session clients went away. With the Windows WinRT API implementation, this is not done separately in the service and so isn't needed in the underlying library code.

### Technical Details

* Move `OnSessionStatusChanged` callback from `UwbDeviceEventCallbacks` to `UwbSessionEventCallbacks` and associated connector code handling tokens and registered callbacks.
* Remove code for processing synchronous IOCTL_UWB_NOTIFICATION I/O (both the simulator and UwbCx process everything async, so this was dead code, making notification processing more complex than it needed to be).
* Changed notification logging to use `std::format`, reducing some verbosity in both the code and the log messages.
* Improve and simplify log messages in `UwbConnector::HandleNotifications`.
* Add debug output appender to Windows `nocli` logging configuration, enabling all logs to be forwarded to an attached debugger, if present.

### Test Results

Ran some ad-hoc scenarios with the simulator driver and verified there were no regressions.

### Reviewer Focus

* Consider whether any of these changes could negatively affect the UwbCx.

### Future Work

* Convert all other Windows-specific logging code to use `std::format` and/or `std::vformat`.
* Simplify token registration code where the template-foo is extremely complex and probably at least 1 layer of indirection can be eliminated (this is inspired by some recent logs that show certain evens are not being raised correctly despite them being delivered by the driver).
 
### Checklist

### DIFF
--- a/windows/tools/nocli/Main.cxx
+++ b/windows/tools/nocli/Main.cxx
@@ -71,13 +71,15 @@ enum class LogInstanceId {
 int
 main(int argc, char* argv[])
 try {
+    static plog::DebugOutputAppender<plog::TxtFormatter> debugAppender{};
     static plog::ColorConsoleAppender<plog::MessageOnlyFormatter> colorConsoleAppender{};
-    plog::init<static_cast<int>(LogInstanceId::Console)>(plog::info, &colorConsoleAppender);
-
     static plog::RollingFileAppender<plog::TxtFormatter> rollingFileAppender(logging::GetLogName("nocli").c_str());
-    plog::init<static_cast<int>(LogInstanceId::File)>(plog::verbose, &rollingFileAppender);
 
-    plog::init(plog::verbose).addAppender(plog::get<static_cast<int>(LogInstanceId::Console)>()).addAppender(plog::get<static_cast<int>(LogInstanceId::File)>());
+    plog::init<static_cast<int>(LogInstanceId::Console)>(plog::verbose, &colorConsoleAppender);
+    plog::init<static_cast<int>(LogInstanceId::File)>(plog::verbose, &rollingFileAppender);
+    plog::init(plog::verbose).addAppender(plog::get<static_cast<int>(LogInstanceId::Console)>())
+        .addAppender(plog::get<static_cast<int>(LogInstanceId::File)>())
+        .addAppender(&debugAppender);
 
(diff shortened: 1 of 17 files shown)


windows/tools/nocli/Main.cxx
```diff
@@ -71,13 +71,15 @@ enum class LogInstanceId {
...
+    static plog::DebugOutputAppender<plog::TxtFormatter> debugAppender{};
     static plog::ColorConsoleAppender<plog::MessageOnlyFormatter> colorConsoleAppender{};
-    plog::init<static_cast<int>(LogInstanceId::Console)>(plog::info, &colorConsoleAppender);
-
     static plog::RollingFileAppender<plog::TxtFormatter> rollingFileAppender(logging::GetLogName("nocli").c_str());
-    plog::init<static_cast<int>(LogInstanceId::File)>(plog::verbose, &rollingFileAppender);
 
-    plog::init(plog::verbose).addAppender(plog::get<static_cast<int>(LogInstanceId::Console)>()).addAppender(plog::get<static_cast<int>(LogInstanceId::File)>());
+    plog::init<static_cast<int>(LogInstanceId::Console)>(plog::verbose, &colorConsoleAppender);
+    plog::init<static_cast<int>(LogInstanceId::File)>(plog::verbose, &rollingFileAppender);
```
corbin-phipps, line 79: What is the reason for adding these two additional `plog::init`s?

corbin-phipps (dismissed):

windows/tools/nocli/Main.cxx
abeltrano: They were there previously, just moved to be declared together with other statics. The one that was added is the `DebugOutputAppender`. That will forward log data to the debugger window, if a debugger is attached. It's really useful when you're using the debugger and want to see things in realtime as opposed to setting up ETW logging separately.

